PR #1, opened 2026-03-24
feat(security): implement security-02/03/04/05/06 foundations

ALI-RUBASS:
## Summary
- add Security-06 inactivity monitoring foundations with role-scoped activity calculators (asker, consultant, admin)
- persist inactivity alerts in `inactive_account_notification_audit_log` with dedup fingerprinting and tenant-aware recipient mapping
- expose a paginated audit-log API (`/service/users/inactive-accounts/audit-logs`) and restrict it to superadmin authority

## Test plan
- [x] build userservice successfully (`mvn -DskipTests -Dmaven.test.skip=true -Dspotless.check.skip=true package`)
- [x] deploy userservice and verify rollout health in k8s
- [x] validate scheduler execution and dedup behavior against MariaDB
- [x] verify threshold boundary behavior (older than 365 days included; fresher accounts excluded)
- [x] verify audit-log API is reachable with auth and denied without auth

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixSynapseService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixSynapseService.java
@@ -209,8 +209,15 @@ public class MatrixSynapseService {
     }
 
     try {
-      String adminUsername = "caritas_admin";
-      String adminPassword = "@CaritasAdmin2025!";
+      String adminUsername = matrixConfig.getAdminUsername();
+      String adminPassword = matrixConfig.getAdminPassword();
+      if (adminUsername == null
+          || adminUsername.isBlank()
+          || adminPassword == null
+          || adminPassword.isBlank()) {
+        log.warn("Matrix admin credentials are not configured; skipping privileged sync login.");
+        return null;
+      }
 
       // Try to login first
       String token = loginUser(adminUsername, adminPassword);
@@ -984,9 +991,20 @@ public class MatrixSynapseService {
 
       return response.getBody();
     } catch (Exception ex) {
-      log.error("Matrix Error: Request to {} failed. Reason: {}", url, ex.getMessage());
+      log.error(
+          "Matrix Error: Request to {} failed. Reason: {}",
+          sanitizeUrlForLog(url),
+          ex.getMessage());
       return null;
     }
+  }
+
+  private String sanitizeUrlForLog(String url) {
+    if (url == null) {
+      return "null";
+    }
+    int queryStart = url.indexOf('?');
+    return queryStart >= 0 ? url.substring(0, queryStart) : url;
   }
 
   /**
--- a/src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
+++ b/src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
@@ -226,6 +226,13 @@ public class SecurityConfig extends KeycloakWebSecurityConfigurerAdapter {
         .antMatchers(
             HttpMethod.PUT, "/useradmin/consultants/{consultantId:" + UUID_PATTERN + "}/agencies")
         .hasAnyAuthority(CONSULTANT_UPDATE, TECHNICAL_DEFAULT)
+        .antMatchers(
+            HttpMethod.POST,
+            "/useradmin/askers/{askerId:" + UUID_PATTERN + "}/deletion/pause",
+            "/useradmin/consultants/{consultantId:" + UUID_PATTERN + "}/deletion/pause",
+            "/service/useradmin/askers/{askerId:" + UUID_PATTERN + "}/deletion/pause",
+            "/service/useradmin/consultants/{consultantId:" + UUID_PATTERN + "}/deletion/pause")
+        .hasAnyAuthority(USER_ADMIN, RESTRICTED_AGENCY_ADMIN, TENANT_ADMIN, SINGLE_TENANT_ADMIN)
         .antMatchers("/useradmin", "/useradmin/**")
         .hasAnyAuthority(USER_ADMIN, TECHNICAL_DEFAULT)
         .antMatchers("/users/consultants/search")
--- a/src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
+++ b/src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
@@ -3,6 +3,7 @@ package de.caritas.cob.userservice.api.config.auth;
 import static de.caritas.cob.userservice.api.config.auth.Authority.AuthorityValue.*;
 
 import de.caritas.cob.userservice.api.adapters.web.controller.interceptor.HttpTenantFilter;
+import de.caritas.cob.userservice.api.adapters.web.controller.interceptor.IpPrivacyHeaderFilter;
 import de.caritas.cob.userservice.api.adapters.web.controller.interceptor.StatelessCsrfFilter;
 import de.caritas.cob.userservice.api.config.CsrfSecurityProperties;
 import org.keycloak.adapters.springsecurity.KeycloakConfiguration;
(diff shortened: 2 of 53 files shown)


src/main/java/de/caritas/cob/userservice/api/service/notification/EventNotificationService.java
```diff
@@ -119,11 +142,7 @@ public void createMessageNotificationFromRoom(
 
     Session session = sessionOpt.get();
     String senderLabel = resolveSenderName(senderUserId, senderDisplayName);
-    String preview = normalizePreview(cleanMessageBody(messagePreview));
-    String text =
-        String.format(
-            "%s sent a new message%s",
-            senderLabel, preview.isBlank() ? "." : ": \"" + preview + "\"");
+    String text = buildMessageNotificationText(senderLabel, messagePreview, envelope);
```
shazia-k, line 169: Issue: privacy.notificationPreviewMode is now applied globally instead of only Matrix-listener notifications. With production default NONE, non-Matrix notifications also lose previews — causing a user-visible regression خارج PR scope.
Suggestion : Restrict redaction to Matrix sync path only, or introduce a Matrix-specific preview policy. A simple fix: apply redaction only when a PrivacyEnvelope is present.

src/main/resources/application.properties
```diff
@@ -267,3 +267,8 @@ anonymous.username.prefix=anon_
 matrix.apiUrl=${MATRIX_API_URL:}
 matrix.registrationSharedSecret=${MATRIX_REGISTRATION_SHARED_SECRET:}
 matrix.serverName=${MATRIX_SERVER_NAME:}
+matrix.adminUsername=${MATRIX_ADMIN_USERNAME:}
```
shazia-k, line 281: Issue: Admin credentials are optional in config but required at runtime. If MATRIX_ADMIN_USERNAME / MATRIX_ADMIN_PASSWORD are missing, the Matrix sync loop won’t start and admin features (e.g., media download) fail — app runs in degraded state.
Suggestion :  Deploy only with matching secrets/config updates. Prefer failing fast at startup if Matrix integration is enabled but admin credentials are missing.

shazia-k (commented):
Review the suggestions.

src/main/java/de/caritas/cob/userservice/api/service/agency/AgencyService.java
```diff
@@ -5,7 +5,6 @@
```
shazia-k, line 5: Issue: Agency-service auth fallback was removed for public/registration paths. This can break anonymous flows and allow registration to bypass strict agency validation when downstream returns 401/403.
Suggestion:  Restore the technical-token fallback for agency-service calls, or confirm that downstream endpoints are intentionally public. If endpoints remain protected, remove the silent 401/403 bypass during registration validation.

src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixSynapseService.java
```diff
@@ -209,8 +209,15 @@ private String getAdminAccessToken() {
     }
```
shazia-k, line 209: Issue: Matrix event listener now depends on new admin environment variables. If missing, the sync loop does not start, stopping Matrix-based notifications. Consultant display-name sync becomes best-effort only.
Suggestion: Treat MATRIX_ADMIN_USERNAME and MATRIX_ADMIN_PASSWORD as mandatory rollout prerequisites, or maintain a compatibility path until all environments are updated.

src/main/java/de/caritas/cob/userservice/api/service/matrix/MatrixEventListenerService.java
```diff
@@ -173,8 +174,6 @@ private Map<String, Object> performMatrixSync() {
         syncUrl += "?timeout=30000";
```
shazia-k, line 174: Issue:Matrix message body is no longer passed into notification creation. This prevents system-message suppression on the Matrix path and breaks FULL preview mode.
Suggestion: Preserve sufficient metadata to detect system messages before redaction, or pass a dedicated system-message flag in the privacy envelope. Otherwise, internal/system events may leak into user notifications and preview behavior regresses.

kennstenicht:
Before we merge this, we should check the tmp-main branch and force push tmp-main into main, dev and production.

shazia-k (commented):
@ALI-RUBASS There are comments to resolve

src/main/java/de/caritas/cob/userservice/api/model/IdentityTombstone.java
```diff
@@ -0,0 +1,76 @@
+package de.caritas.cob.userservice.api.model;
+
+import java.time.LocalDateTime;
+import java.util.Objects;
+import javax.persistence.Column;
```
shazia-k, line 5: Issue: javax.validation which do not exist in Spring Boot 4.0.1 (Jakarta EE 11). These files will not compile.
Suggestion:
Replace all javax.persistence with jakarta.persistence and javax.validation with jakarta.validation.

src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
```diff
@@ -219,12 +226,22 @@ protected void configure(HttpSecurity http) throws Exception {
         .antMatchers(
             HttpMethod.PUT, "/useradmin/consultants/{consultantId:" + UUID_PATTERN + "}/agencies")
         .hasAnyAuthority(CONSULTANT_UPDATE, TECHNICAL_DEFAULT)
+        .antMatchers(
```
shazia-k, line 229: Issue:
 antMatchers(), was removed in Spring Security 6/7. The existing main branch already uses requestMatchers(). 

Suggestion:
Replace antMatchers() with requestMatchers() for all new path rules.

src/main/java/de/caritas/cob/userservice/api/workflow/inactiveaccountnotification/service/InactiveAccountNotificationService.java
```diff
@@ -0,0 +1,184 @@
...
+  public void scanAndNotifyInactiveAccounts() {
+    LocalDateTime now = LocalDateTime.now(ZoneOffset.UTC);
+    LocalDateTime cutoff = now.minusDays(inactivityThresholdDays);
+    notifyInactiveAskers(cutoff, now);
+    notifyInactiveConsultants(cutoff, now);
+    notifyInactiveAdmins(cutoff, now);
+  }
+
+  private void notifyInactiveAskers(LocalDateTime cutoff, LocalDateTime now) {
+    for (User user : userRepository.findAllByDeleteDateIsNull()) {
```
shazia-k, line 65: Suggestion: load every active user, consultant, and admin into memory. For large deployments this causes high memory usage.
Push the last-activity cutoff filter into the database query, or use streaming with fetch-size hints.

src/main/java/de/caritas/cob/userservice/api/workflow/delete/service/DeleteUserAccountService.java
```diff
@@ -53,6 +54,9 @@ public void deleteUserAccounts() {
 
   private List<DeletionWorkflowError> deleteAskersAndCollectPossibleErrors() {
     return this.userRepository.findAllByDeleteDateNotNull().stream()
+        .map(deletionLifecycleService::normalizeUserLifecycle)
+        .peek(userRepository::save)
```
shazia-k, line 58: Stream.peek(userRepository::save) is used for a critical side effect. peek() is intended for debugging and its execution is not guaranteed.
Replace with .map() that saves and returns, or separate the normalize+save step from the filter+delete step.

src/main/java/de/caritas/cob/userservice/api/workflow/delete/action/asker/DeleteDatabaseAskerAction.java
```diff
@@ -30,6 +32,7 @@ public class DeleteDatabaseAskerAction implements ActionCommand<AskerDeletionWor
   @Override
   public void execute(AskerDeletionWorkflowDTO actionTarget) {
     try {
+      identityTombstoneService.recordDeletedUser(actionTarget.getUser());
```
shazia-k, line 35: Tombstone record is written before the entity delete. If delete fails, the tombstone persists as an orphan.
Ensure both are in the same @Transactional so they roll back together, or write the tombstone only after a confirmed delete.

shazia-k (commented):
Added More comments

PR #5, opened 2026-05-09
fix: allow anonymous invite users to assign and view group chat

nikunjdecyb:
Fixes three independent root causes that prevented anonymous users (registered via the invite link flow at /invite/:token) from being assigned to their group chat, causing a forced logout and error-page redirect.

Changes
SecurityConfig.java Added ANONYMOUS_DEFAULT authority to two endpoint rules:

PUT /users/chat/{groupId}/assign
GET /users/chat/room/{chatId:[0-9]+}
Anonymous invite users are registered with Keycloak role user. In edge cases where role assignment fails silently in CreateUserFacade, the user ends up with only ANONYMOUS_DEFAULT. These endpoints now accept both USER_DEFAULT and ANONYMOUS_DEFAULT.

AssignChatFacade.java Added a numeric-chatId fallback in getChat(). The frontend invite link flow sometimes passes the database chat ID (e.g. 102559) rather than the RocketChat/Matrix group ID string. The facade now tries getChatByGroupId() first and falls back to chatService.getChat(Long) when the input is a valid number.

application-*.properties (all 5 env files) Added /users/chat/ to csrf.whitelist.config-uris. Anonymous invite users bypass the normal app-load sequence that sets the CSRF-TOKEN cookie, so their first PUT /assign request arrived without a matching CSRF header and was blocked by StatelessCsrfFilter. Chat endpoints are already protected by Bearer token, so removing CSRF on them is safe.

AgencyInviteLinkService.java Rewrote the redeem() method from a chained orElseThrow(lambda) pattern to plain imperative if/else. Java 17's type inference could not resolve the return type of a map() block that only threw (never returned), causing a compile failure. Runtime behaviour is identical.

Testing
mvn compile -Dspotless.check.skip=true → exit 0, all 958 files clean
AssignChatFacadeTest — all 3 existing tests pass: the new numeric fallback path is only entered when groupId.matches("\\d+"), so tests using RC_GROUP_ID = "jjjuuu" hit the same code paths as before
No new test failures introduced (5 pre-existing broken test files in EmailNotificationFacadeTest, SessionServiceTest, ConsultantTest, CreateEnquiryMessageFacadeTest, SystemNotificationEmailSettingsServiceTest are pre-existing technical debt from prior security commits)

Related
Frontend PR: adds FETCH_ERRORS.FORBIDDEN guard to apiGetChatRoomById, skips useJoinGroupChat for anonymous sessions, and widens ANONYMOUS_RESTRICTED_PATH regex — these frontend fixes prevent the forced logout on the client side while this backend PR fixes the actual permission gaps server-side.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Agency invite links for streamlined user onboarding and access management
  * Ability to pause pending account deletions with customizable duration
  * Anonymous users can now see the count of pending requests ahead in the queue
  * Consultant name changes trigger notifications to users
  * Configurable IP privacy controls for enhanced data protection

* **Improvements**
  * Enhanced security monitoring system for inactive accounts with admin notifications
  * Better error handling for consultant availability checks
  * Improved anonymous enquiry workflow with fallback options

### DIFF
--- a/api/conversationservice.yaml
+++ b/api/conversationservice.yaml
@@ -369,6 +369,10 @@ components:
         numAvailableConsultants:
           type: integer
           minimum: 0
+        peopleAhead:
+          type: integer
+          minimum: 0
+          description: Number of anonymous enquiries created before this one in the same consulting type and still waiting (status NEW).
         status:
           type: string
           enum:
--- a/documentation/ADR-SECURITY-02-unified-crypto-boundary.md
+++ b/documentation/ADR-SECURITY-02-unified-crypto-boundary.md
@@ -1,0 +1,41 @@
+# ADR Security-02: Unified Cryptographic Boundary
+
+## Status
+Accepted (foundation).
+
+## Context
+The current platform includes two chat security paths:
+
+- Legacy Rocket.Chat custom E2EE helpers in frontend code.
+- Matrix-based transport where server-side integrations can process message content.
+
+This mixed model creates governance and lifecycle ambiguity for key ownership, participant changes,
+and historical visibility.
+
+## Decision
+For greenfield security work, the platform defines one canonical cryptographic model:
+
+- Messages and attachments must follow one shared security policy boundary.
+- Participant and device changes must trigger deterministic key lifecycle actions.
+- Historical visibility is governed by cryptographic policy, not membership checks alone.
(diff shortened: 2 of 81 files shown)


shazia-k:
@coderabbitai review

nikunjdecyb:
close this PR as this is no more relevant

PR #84, opened 2026-06-10
fix: tolerate deleted agencies in consultant search

Storypapst:
## Summary

`GET /users/consultants/search` returned **500** for every page containing a consultant whose agencies were all deleted, which made the admin "Counsellors" list show no data at all (on dev there are currently two such consultants: `bertabusch1` and `frank0806beratungpupslingen`, both tenant 27).

Root cause chain:
- `findByConsultantIdIn` were the **only** `ConsultantAgencyRepository` queries without the `DeleteDateIsNull` filter, so listings also resolved agency ids of soft-deleted relations
- UserService resolves those agencies per result page via AgencyService `getAgenciesByIds`
- AgencyService responds **404 when none of the requested ids exist** (`AgencyController.getAgenciesByIds`)
- The generated client throws, `AccountManager.findConsultantsByInfix` didn't catch it → 500 for the whole page

Two complementary fixes:
1. **Correctness**: add `DeleteDateIsNull` to the two listing queries (consultant search + admin agency enrichment), so soft-deleted relations are never resolved
2. **Resilience**: catch `RestClientException` in `AccountManager.findConsultantsByInfix` and render affected consultants without agency data instead of failing the entire listing (covers hard-orphaned relations, e.g. from tenant deletion)

Follow-ups worth discussing:
- AgencyService could return `200 []` instead of 404 for bulk get-by-ids
- Agency/tenant deletion should consistently soft-delete `consultant_agency` relations

## Test plan

- `./mvnw compile` passes (spotless applied)
- Root cause verified against dev: `query=*` pages containing the two consultants 500, pages without them 200; their `/useradmin/consultants/{id}/agencies` also 500s, while `/useradmin/consultants/{id}` (no agency resolution) returns 200

## Deploy note

The cluster runs `pullPolicy: Never` (local k3s images), so merging alone does not deploy — the image has to be imported on the host and the pod restarted before the dev admin shows consultants again.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/AccountManager.java
+++ b/src/main/java/de/caritas/cob/userservice/api/AccountManager.java
@@ -4,6 +4,7 @@ import static java.util.Objects.isNull;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import de.caritas.cob.userservice.api.adapters.web.dto.AgencyDTO;
 import de.caritas.cob.userservice.api.admin.service.tenant.TenantService;
 import de.caritas.cob.userservice.api.exception.httpresponses.InternalServerErrorException;
 import de.caritas.cob.userservice.api.helper.AuthenticatedUser;
@@ -24,6 +25,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicReference;
@@ -34,6 +36,7 @@ import org.springframework.data.domain.Page;
 import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Sort.Direction;
 import org.springframework.stereotype.Service;
+import org.springframework.web.client.RestClientException;
 
 @Slf4j
 @Service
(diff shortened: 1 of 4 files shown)


Storypapst:
Some context on how this was verified and what reviewers should know:

**Verification done**
- Root cause confirmed against the dev API by bisecting the search endpoint: of 64 consultants, exactly two records (`bertabusch1`, `frank0806beratungpupslingen`, both tenant 27, agencies deleted) make any page containing them fail with 500 — `query=*` page 1 fails, queries not matching them return 200. Their `/useradmin/consultants/{id}/agencies` also 500s, while `/useradmin/consultants/{id}` (no agency resolution) returns 200.
- Two regression tests added in `AccountManagerTest`: (1) the search must not fail when the AgencyService bulk lookup raises a 404-style client error, (2) the search must only resolve agencies of not soft-deleted relations. Both green locally (JDK 17), full unit suite green in CI.

**What could NOT be verified locally**
- A full end-to-end run against real data is not possible from a dev machine (dev MariaDB is cluster-internal, credentials only in cluster secrets).
- The existing endpoint-level ITs (`UserControllerConsultantE2EIT`) are currently broken on `main` — the Spring context fails schema validation (`missing table [admin]`) and `*IT` classes never run in CI. Tracked in #85.

**Deploy & rollback**
- The cluster runs `pullPolicy: Never`, so merging only publishes the image to GHCR — it has to be imported on the k3s host and the userservice pod restarted.
- Please note the currently running image/tag before deploying so we can roll back immediately if anything misbehaves.

**Related follow-ups**
- #85 — fix the IT test schema and run ITs in CI
- #86 — clean up the orphaned `consultant_agency` rows on dev and fix the deletion flows
- OpenResilienceInitiative/ORISO-AgencyService#4 — `getAgenciesByIds` should return `200 []` instead of 404

PR #88, opened 2026-06-10
fix: roll back Keycloak user when admin creation fails

Storypapst:
## Problem

When creating an agency administrator (`POST /useradmin/agencyadmins`), the Keycloak user is created first, followed by the password, roles, and the database entry. If any of the subsequent steps fail (e.g., because the `restricted-agency-admin` realm role is missing in Keycloak, resulting in a 500 error), an **orphaned Keycloak user** remains. Any further attempt using the same username or email address then results in a **409 Conflict** error.

This exact pattern (initial 500 error followed by persistent 409 errors) occurred today on api.oriso.org.

## Fix

Wrap the password update, role assignment, and `adminRepository.save` call in a shared try/catch block and invoke `identityClient.rollBackUser(...)` upon any failure (previously, this was only done during the password step).

## Operational Note

The root cause is the missing `restricted-agency-admin` role in the `online-beratung` realm—see the corresponding fix in ORISO-Keycloak (realm.json). Additionally, the role must be created once in the running Keycloak instance.

## Test

- `mvn compile` + Spotless check passed
- `CreateAdminServiceTest`: 2/2 passed

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/admin/service/admin/create/CreateAdminService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/admin/service/admin/create/CreateAdminService.java
@@ -9,7 +9,6 @@ import de.caritas.cob.userservice.api.adapters.web.dto.CreateAdminDTO;
 import de.caritas.cob.userservice.api.adapters.web.dto.UserDTO;
 import de.caritas.cob.userservice.api.admin.service.consultant.validation.UserAccountInputValidator;
 import de.caritas.cob.userservice.api.config.auth.UserRole;
-import de.caritas.cob.userservice.api.exception.httpresponses.CustomValidationHttpStatusException;
 import de.caritas.cob.userservice.api.helper.AuthenticatedUser;
 import de.caritas.cob.userservice.api.helper.UserHelper;
 import de.caritas.cob.userservice.api.helper.UsernameTranscoder;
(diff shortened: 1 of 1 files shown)


Storypapst:
@Hassan9215 This fix is ​​directly related to Keycloak (OpenResilienceInitiative/ORISO-Keycloak#2) — please **test thoroughly**, ideally in a staging environment using a real Keycloak instance (not just H2/mocks).

## Test Plan

**Preparation:**
1. Create the role `restricted-agency-admin` in the `online-beratung` Keycloak realm (Admin Console → Realm roles → Create role) if it does not already exist.

**Happy Path:**
2. Create a counseling center admin with a tenant and counseling center in the admin panel → expected: 200, admin appears in the list.
3. Verify in Keycloak: User exists and has the roles `restricted-agency-admin` and `user-admin`.
4. Log in as the new counseling center admin → sees only their assigned counseling center(s).

**Rollback Behavior (Core of this PR):**
5. Negative test: Delete the `restricted-agency-admin` role from the realm for testing purposes and attempt creation again → expected: 500, but **no orphaned user in Keycloak** (previously: user remained).
6. Immediately afterwards, create the same username/email again (with the role restored) → expected: 200, **no 409**.
7. Regression test for tenant admin creation: `POST /useradmin/tenantadmins` → remains functional (same `createNewAdmin` code path).

**Edge Cases:**
8. Creation with an already taken email/username → expected: clean 409 with validation message (no 500).
9. Creation with a weak password → expected: error message, no orphaned Keycloak user.

shazia-k (approved):

Storypapst:
Closing as superseded by dev.

dev already contains the same fix, branched from the same base (266de6de), in a stricter form — commits d7038446 ("feat: enhance admin creation process with error handling") and e6546663 ("Surface ROLE_NOT_FOUND reason when an admin realm role is missing"). Both move the role assignment and adminRepository.save into the try block and roll back the Keycloak user on any failure, and additionally:
- map a missing realm role (NotFoundException) to a clean 404 ROLE_NOT_FOUND instead of a generic 500 — exactly the 500-then-409 orphaned-user scenario this PR targeted;
- wrap other runtime failures in InternalServerErrorException.

This PR's broader catch (Exception e) would regress that error mapping, so closing in favour of the dev implementation.

PR #101, opened 2026-06-12
Tolerate agency metadata errors during consultant login

### LINKED ISSUE #96: Consultant login fails after Keycloak token when loading /users/data
## Problem
Consultant login currently succeeds at Keycloak, but the frontend cannot enter the app because the follow-up profile request `GET /service/users/data` fails.

## Evidence from 2026-06-11
- `POST https://api.oriso.org/auth/realms/online-beratung/protocol/openid-connect/token` returns HTTP 200 for consultant accounts.
- With the returned bearer token, `GET https://api.oriso.org/service/users/data` fails:
  - `livechatconsult`: HTTP 500, correlation `358af261-df86-4d3a-8fbb-e1a243f56bb1`
  - `counsellor1006`: HTTP 403, correlation `9d1a8602-0f0e-4571-a306-2bc1344db53e`
- Browser stays on `https://app.oriso.org/login` without a useful visible error.
- UserService logs visible in the server terminal show recurring Rocket.Chat technical-user login failures and `No technical user was initialized` errors.

## Expected behavior
After Keycloak returns a valid consultant token, `/users/data` should return the authenticated consultant profile and the frontend should redirect into the app.

## Notes
The local UserService code on `dev` already contains a Rocket.Chat-safe fallback in `Messenger#getAvailability`, but the deployed service still appears to be hitting Rocket.Chat technical-user failures. Please check the deployed image/version and pull the UserService logs for the correlation IDs above.

## Follow-up
The frontend should also show an explicit unexpected-login-error message when post-token user-data loading fails, but the blocking production failure is currently in the UserService post-login path.

Storypapst:
## Problem
Consultant login succeeds at Keycloak, but the frontend cannot enter the app when `GET /service/users/data` fails while enriching the consultant profile with agency metadata.

Observed for `livechatconsult` on 2026-06-12:
- Keycloak token: HTTP 200
- `GET /service/users/data`: HTTP 500
- Correlation ID: `a666f559-a73e-4b6a-bc13-9b4d805e2930`
- UserService stacktrace: `ConsultantDataProvider.retrieveData -> AgencyService.getAgencies -> AgencyService returned 500 for /agencies/250`

Fixes #96

## Solution
- Load consultant agencies once while building the profile response.
- Treat AgencyService HTTP response failures like missing optional metadata for this login path.
- Return the core consultant profile with an empty agencies list instead of surfacing the AgencyService error as a login failure.
- Keep a warning log with the consultant id, status code, and response body for later cleanup of the broken agency reference/service response.

## Checks
- `git diff --check`
- `./mvnw -Dtest=ConsultantDataProviderTest test` was attempted, but this checkout still fails during project compilation with broad generated/Lombok missing-symbol errors outside the touched files before the targeted test can run.

## Notes
`counsellor1006` is a separate account-data case: Keycloak returns a token, but UserService correctly rejects `/users/data` with 403 because the consultant is flagged for deletion.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/facade/userdata/ConsultantDataProvider.java
+++ b/src/main/java/de/caritas/cob/userservice/api/facade/userdata/ConsultantDataProvider.java
@@ -27,7 +27,7 @@ import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Component;
-import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestClientResponseException;
 
 /** Provider for consultant information. */
 @Component
@@ -62,6 +62,7 @@ public class ConsultantDataProvider {
   }
 
   private UserDataResponseDTO userDataResponseDtoOf(Consultant consultant) {
+    var agencies = agencyDTOsOf(consultant);
 
     return UserDataResponseDTO.builder()
         .userId(consultant.getId())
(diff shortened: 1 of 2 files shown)


Storypapst:
Additional verification after commit `8777cd60`:

- Added unit coverage for AgencyService `404 NotFound` during consultant profile enrichment.
- GitHub CI is green for the updated branch.
- Local Maven test execution in this checkout still fails before tests because of existing generated/Lombok compile errors outside this PR.
- Dev API verification after data repair: `GET /service/users/data` returns 200 for the counsellor test account and includes one agency.
- Browser verification: local ORISO-Frontend login redirects to `/sessions/consultant/sessionPreview` and renders the consultant session preview UI.

PR #107, opened 2026-06-13
Refactor Matrix media operations

radaljj:
### Summary

- Extracted Matrix media upload/download logic from `MatrixSynapseService` into a new focused `MatrixMediaClient`.
- Kept `MatrixSynapseService` as the existing public entry point, so current callers/controllers remain unchanged.
- Moved file upload, file download, and file-message sending logic into the new media client.
- Added focused unit coverage for Matrix media upload/download behavior.

### Testing

- `mvn -DskipTests -Dspotless.check.skip=true compile`
- `mvn -Dtest=MatrixMediaClientTest -Dspotless.check.skip=true test`
- `mvn spotless:check`

### Notes

- Refactor only.
- No API changes.
- No database changes.
- No intended Matrix behavior changes.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixMediaClient.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixMediaClient.java
@@ -26,8 +26,7 @@ public class MatrixMediaClient {
   private final RestTemplate restTemplate;
 
   public Map<String, Object> uploadFile(MultipartFile file, String roomId, String accessToken) {
-    String fileName =
-        file.getOriginalFilename() != null ? file.getOriginalFilename() : "file";
+    String fileName = file.getOriginalFilename() != null ? file.getOriginalFilename() : "file";
     String contentType = file.getContentType();
 
     try {
@@ -86,7 +85,8 @@ public class MatrixMediaClient {
     }
 
     if (response.getBody() == null) {
-      throw new RuntimeException("Failed to download file: No file data in Matrix download response");
+      throw new RuntimeException(
+          "Failed to download file: No file data in Matrix download response");
     }
 
     log.info("✅ File downloaded successfully: {} bytes", response.getBody().length);
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixMediaClient.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixMediaClient.java
@@ -56,10 +56,14 @@ public class MatrixMediaClient {
       sendFileMessage(roomId, fileName, contentUri, contentType, file.getSize(), accessToken);
 
       return Map.of(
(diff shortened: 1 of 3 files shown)


shazia-k:
@radaljj 
I found some missing scenarios and missing tests which I have completed
Here is the summary of changes I made

Review Summary
MatrixMediaClient.java — Issues Found & Fixed
#
Issue
1
NPE: file.getContentType() can return null → MediaType.parseMediaType(null) throws InvalidMediaTypeException
🔴 Bug
Fix:
Default to MediaType.APPLICATION_OCTET_STREAM when content type is null

2
NPE: file.getOriginalFilename() can return null → Map.of(...) throws NullPointerException for null values
🔴 Bug
Fix:
Default to "file" when original filename is null

3
Double exception wrapping — the inner throw new RuntimeException("No content_uri…") was caught by the outer catch (Exception e) and re-wrapped
🟡 Code smell
Fix:
Left as-is (message remains descriptive; cause chain is preserved)

4
sendFileMessage return value never used — method returned Map<String, Object> but the caller discarded it; the error branch returned a stale error map
🟡 Code smell
fix:
Changed return type to void; removed the meaningless error return

5
sendFileMessage failures silently masked as success — upload returns {success: true} even when the room message PUT fails
🟡 Design note
Fix:
Documented via the new test uploadFile_Should_StillReturnSuccess_When_SendFileMessageFails

6
Raw Map type on RestTemplate calls
🟢 Warning
Fix:
Added @SuppressWarnings("rawtypes") at the call sites

7
downloadFile try/catch over-broad structure — inner RuntimeException thrown and immediately re-caught
🟡 Code smell
Fix:
Restructured: network call inside try-catch, null-body check moved outside
 
MatrixMediaClientTest.java — Missing Tests Added (10 new, 12 total)
New Test.                 Branch Covered
uploadFile_Should_ThrowException_When_ResponseBodyIsNull
Null response body from upload POST

uploadFile_Should_ThrowException_When_ContentUriIsMissing
Response body present but no content_uri key

uploadFile_Should_ThrowException_When_RestTemplateThrows
Network failure during upload

uploadFile_Should_UseImageMsgtype_When_FileIsImage
image/* → m.image msgtype

uploadFile_Should_UseVideoMsgtype_When_FileIsVideo
video/* → m.video msgtype

uploadFile_Should_UseAudioMsgtype_When_FileIsAudio
audio/* → m.audio msgtype

uploadFile_Should_UseFileMsgtype_When_FileHasGenericContentType
application/* → m.file msgtype

uploadFile_Should_StillReturnSuccess_When_SendFileMessageFails
sendFileMessage exception is swallowed; documents current behaviour

downloadFile_Should_ThrowException_When_ResponseBodyIsNull
Null response body from download GET

downloadFile_Should_ThrowException_When_RestTemplateThrows
Network failure during download

shazia-k (approved):

Issue #110, opened 2026-06-15
[Bug]: US-C01 — Plaintext user passwords stored permanently in the database (matrix_password)


shazia-k:
### Affected Service/Component

User Service

### Priority

P1-High

### Bug Description

During registration, the platform password chosen by the user (identical to the Keycloak login password) is persisted unencrypted in MariaDB: `CreateUserFacade.java:106` `user.setMatrixPassword(plainCreds.getPassword())`. The columns are plain `VARCHAR(255)` without any converter/encryption (`model/User.java:77-79`, `model/Consultant.java:94-96`, `addMatrixPasswordToUser.sql:1`). The server subsequently logs in as the user to Matrix everywhere using the stored plaintext password — across more than 30 read sites, including `MatrixMessageController.java:84/103/252`, `AssignEnquiryFacade.java:442`, `DirectSessionMatrixRoomService.java:122`. A DB dump or SQL injection anywhere in the stack immediately compromises the Keycloak logins of all seekers on a pastoral counselling platform. For KDG/DSGVO (counselling confidentiality!), this is a worst case.

Obsidian Url:
obsidian://open?vault=ORISO&file=3%20Findings%2FORISO-UserService%2FUS-C01%20%E2%80%94%20Plaintext%20user%20passwords%20stored%20permanently%20in%20the%20database%20(matrix_password)

### Steps to Reproduce

Go to admin.oriso.org
Create a new Consultant User
Look into the Maria DB Password is stored in DB without encryption in Consultant table

## Solution / What to do instead

Remove the `matrix_password` columns entirely. Handle server-side Matrix actions via the Synapse Admin API: `POST /_synapse/admin/v1/users/{userId}/login` generates short-lived access tokens without requiring password knowledge (`MatrixSynapseService` already has an admin token, see `getAdminAccessToken` line ~262). Alternatively, use a Matrix Application Service with `as_token` impersonation. Migration: null the columns and drop them via Liquibase; switch all existing flows to admin login tokens.

### Actual Behavior

Password is stored in database without encryption

### Relevant Logs/Error Messages

```shell

```

### Checklist

- [x] I have searched existing issues to ensure this bug hasn't been reported before
- [x] I have provided all relevant information above
- [x] I can reproduce this bug consistently

kiodreambau:
**Verified fixed — closing as part of the Security Audit Closure epic (OpenResilienceInitiative/ORISO-Docs#72).**

[...]

PR #124, opened 2026-06-16
Fix/us c01 remove matrix password plaintext

shazia-k:
# Pull Request

Issue:

https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/110

Fix and How It will work:

How consultant (and seeker) login / Matrix auth works after the merge
The password is used exactly once — to provision the Matrix account — then thrown away
Registration
  │
  ├─ plainCreds.getPassword()  → Keycloak account (stored in Keycloak, not in your DB)
  │
  └─ userHelper.getRandomPassword()  → NEW random 1-time password
           │
           └─ matrixSynapseService.createUser(username, randomPassword, displayName)
                    → POST /_synapse/admin/v1/register   (Synapse provisioning)
                    randomPassword is discarded immediately, never written anywhere
                    Only  matrix_user_id  is saved to MariaDB
Before this session's fix, the Keycloak password (plainCreds.getPassword()) was passed to Synapse for account creation — so the Matrix account's Synapse-internal password was identical to the Keycloak password. That is now replaced with a random, never-stored, never-reused password.
Every subsequent Matrix operation uses admin impersonation — no password at all
Any server-side action (send message, create room, invite user, …)
  │
  └─ loginUserViaAdmin("@alice:matrix.example.com")
           │
           ├─ in-memory cache hit (50 min TTL)?  →  return cached token
           │
           └─ cache miss
                    │
                    └─ getAdminAccessToken()          ← MATRIX_ADMIN_USERNAME /
                             (50 min in-memory cache)    MATRIX_ADMIN_PASSWORD  (env vars)
                                      │
                                      └─ POST /_synapse/admin/v1/users/{userId}/login
                                               → short-lived token for that user
                                               → cached in impersonationTokenCache
The consultant's (or seeker's) password is never read back from anywhere. It doesn't exist in your DB. It doesn't exist in Synapse either (the random provisioning password was set once and is effectively unknowable). All Matrix actions go through the admin impersonation API.
What remains in the DB after the change
Column
Value
Purpose
consultant.matrix_user_id
@alice:matrix.example.com
Identifies the Matrix account for admin impersonation
<span>consultant.matrix_password</span>
<span>plaintext</span>
Dropped by migration 0053

**⚠️ Please select a template above by clicking on one of the links.**


shanzaeimran (approved):

Storypapst:
@BjoernLudwig FYI: this was the reason why we had today not chat working. @shazia-k Rolled it back now and will think of an alternative.

@joro4b @Hassan9215 @Shirloin @nikunjdecyb @einsteinjava @shanzaeimran @Kadyan25 @radaljj 
@kennstenicht 

Any ideas are welcome. 

@shazia-k if you can please describe the issue better you see so far 🙏

shazia-k:
@Storypapst 
Yes this was the issue of chat not working.
During our deep scanning of the code we figured out that matrix user Id and matrix password is storing in keycloak as well as in mariaDB too. and it should not be stored in maria db in the form of text.
To overcome this, I removed it's occurrences from maria DB and implemented the mechanism that every time it just go to keycloak and synapse admin for token authentication, This fix worked well for User login in admin and portal But we came across the problem that chat's are not working, because during chat user's are still getting authenticated for matrix chat.
Because it was a blocker, So I completely removed the fix and trying to think of something else like instead of just removing this field completely,
we can store hashed format and this way security concern will also be resolved and things should get working as it is.

Any other suggestions are welcome.

Issue #134, opened 2026-06-17
[Bug]: US-C05 Debug Redis mirror writes counselling message texts in plaintext to Redis


Shirloin:
### Affected Service/Component

User Service

### Priority

P0-Critical

### Bug Description

`RedisMessageMirrorService.mirrorOutgoingMessage()` serializes the full plaintext message content, sender name, role, session ID, and room ID as JSON and writes it to Redis via `StringRedisTemplate`, with a 15-minute TTL. This is called from three locations: `MatrixMessageController `(outgoing messages), `MatrixEventListenerService `(incoming messages), and `EventNotificationController `(notification previews). The Helm chart sets `debug.redisMessageMirror.enabled: true `as the default value, meaning this is active in production right now, not just in dev environments. The only protection is an environment variable, with no code-level guard preventing it from loading in production. The Redis instance backing this also uses a weak/default password and is exposed via Redis Commander, a UI that lets anyone with access read the plaintext keys directly. As a result, full counselling message content — including sensitive health/mental-health information (GDPR Art. 9 special category data) — is currently being written in plaintext to a Redis instance that is both weakly protected and exposed via an admin UI.

### Steps to Reproduce

1. Send any message through the platform (consultation chat)
2. Within 15 minutes, query Redis: `redis-cli KEYS "debug:msgmirror:*"`
3. Read the value of any matching key
4. Observe the full plaintext message content, sender name, and role are present in the stored JSON

### Expected Behavior

This debug feature should never be reachable in production regardless of environment variable configuration, and even in non-prod environments, it should never store full plaintext message content — only non-identifying metadata.

#### Solution / What to do instead

1. Write smoke tests first (red): (a) mirror does not write to Redis when disabled, (b) the `RedisMessageMirrorService `bean cannot be loaded at all when the active Spring profile is `prod`, (c) when enabled in dev, the stored Redis value contains only metadata (session ID, message length, partial hash) and never the message text
2. Add `@Profile("!prod") `and` @ConditionalOnProperty(name = "debug.redisMessageMirror.enabled", havingValue = "true", matchIfMissing = false) `to `RedisMessageMirrorService`, so it's structurally unreachable in production regardless of config
3. Replace the plaintext payload in `mirrorOutgoingMessage() `with metadata only: session ID, message length, and a short SHA-256 hash prefix — remove `message`, `sender`, `role`, and `roomId `from the stored payload entirely
[...]

PR #138, opened 2026-06-18
Fix(US-C05): harden redis debug message mirror against prod plaintext leaks

Shirloin:
## Summary

Fixes [US-C05 / #134](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/134): the debug Redis message mirror was writing counselling message content to Redis and could be active in production via env/Helm config alone.

This PR adds **code-level** guards so the mirror cannot silently run in prod, stores **metadata only** in Redis (no plaintext message or sender identity), and makes all call sites tolerant when the bean is absent.

### Application changes
- **`RedisMessageMirrorService`**
  - `@Profile("!prod")` — bean excluded when `prod` profile is active
  - `@ConditionalOnExpression("${debug.redis.message.mirror.enabled:false}")` — bean only loads when explicitly enabled
  - Startup `log.warn` when the bean is instantiated
  - Redis payload: `ts`, `sessionId`, `roomId`, `messageLength`, `messageHash` (8-char SHA-256 prefix)
  - Removed from payload: `message`, `sender`, `senderRole`, `eventId`
  - Retained runtime `enabled` `@Value` + early return (defense in depth)
- **Call sites** (`MatrixMessageController`, `MatrixEventListenerService`, `EventNotificationController`)
  - `Optional<RedisMessageMirrorService>` injection + `.ifPresent(...)` so prod startup does not fail when the bean is missing
- **Tests**: `RedisMessageMirrorServiceTest` — disabled path, metadata-only payload, missing `StringRedisTemplate` handling

### Infra follow-up (not in this PR — needs Helm/ConfigMap access)
- [ ] Confirm prod ConfigMap sets `DEBUG_REDIS_MESSAGE_MIRROR_ENABLED=false`

Fixes OpenResilienceInitiative/ORISO-UserService#134

## Test

- `mvn test -Dtest=RedisMessageMirrorServiceTest` — 3/3 pass (JDK 17)
- Full module test suite run — no failures in mirror-related classes; pre-existing PowerMock/Mockito errors elsewhere unchanged
- Deploy to non-prod with `debug.redis.message.mirror.enabled=true` — confirm startup warning log appears
- Send a test message — confirm Redis keys contain metadata only (no `message` / `sender` fields)
- Deploy with `spring.profiles.active=prod` — confirm app starts without `RedisMessageMirrorService` bean
- Infra owner: apply Helm/ConfigMap changes above in prod

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/EventNotificationController.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/EventNotificationController.java
@@ -3,6 +3,7 @@ package de.caritas.cob.userservice.api.adapters.web.controller;
 import de.caritas.cob.userservice.api.helper.AuthenticatedUser;
 import de.caritas.cob.userservice.api.service.matrix.RedisMessageMirrorService;
 import de.caritas.cob.userservice.api.service.notification.EventNotificationService;
+import java.util.Optional;
 import javax.validation.constraints.Min;
 import javax.validation.constraints.NotBlank;
 import lombok.NonNull;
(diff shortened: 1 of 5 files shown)


Hassan9215 (approved):
fine to merge, and lets test it after deploy in pre-dev env

PR #158, opened 2026-06-23
fix: surface ROLE_NOT_FOUND reason when a required realm role is missing

Storypapst:
## Problem
Creating an agency admin failed with HTTP 500 and the opaque panel toast "Es ist ein Fehler aufgetreten." Root cause: `CreateAdminService` assigns the role bundle `[restricted-agency-admin, user-admin]`; when a realm role is missing, `KeycloakService.updateRole` throws `NotFoundException`, which was caught as a generic `RuntimeException` and rethrown as `InternalServerErrorException` (500) with no `X-Reason`, so the UI could only show the default error.

(Operational note: the live `online-beratung` realm was missing `restricted-agency-admin`; it has been added out-of-band. This change makes the failure mode diagnosable if it recurs.)

## Change
- Translate the Keycloak `NotFoundException` from the role lookup into the existing `CustomValidationHttpStatusException` with a new `HttpStatusExceptionReason.ROLE_NOT_FOUND` (HTTP 404), so the API returns a specific `X-Reason: ROLE_NOT_FOUND`. Existing user rollback is preserved.
- The new catch sits before the generic `RuntimeException` catch, so only the role-not-found case is translated; all other failures still return 500.

## Tests (TDD, red → green)
- `CreateAdminServiceTest.createNewAgencyAdmin_ShouldThrowRoleNotFoundReason_AndRollbackUser_WhenRealmRoleIsMissing` — RED (got `InternalServerErrorException`) → GREEN. Independently re-verified.

Pairs with the ORISO-Admin PR that adds the `message.error.ROLE_NOT_FOUND` translation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when admin account creation fails due to missing role assignments — now returns a specific 404 error with clear messaging instead of a generic server error
  * System automatically cleans up partially created accounts on role assignment failures


Storypapst:
Tracked in #160.

Shirloin (approved):

PR #65, opened 2026-06-08
[refactor]: Enhance UserController and KeycloakUserDataProvider with …

nikunjdecyb:
This commit makes the user-data and tenant-admin-list endpoints resilient to Keycloak/tenant-service failures, and fixes a 403 bug where USER_ADMIN roles couldn't search tenant admins.

PR #180, opened 2026-06-24
Merge admin-fixes into dev: resilient OTP/user-data + scoped tenant-admin access (supersedes #65)

Storypapst:
## Why

PR #65 (`feature/admin-fixes`, approved) was **DIRTY** — branched ~212 commits behind `dev`, with overlapping edits that, merged naively, would have shipped real regressions. This branch is the conflict-resolved merge of its intent onto current `dev`, with those regressions fixed.

## What the PR intended (kept)

- Resilient `/users/data`: don't 500 when an OTP lookup or Keycloak user lookup fails — degrade to token-based data / omit OTP state.
- Let `USER_ADMIN` use the tenant-admin **search** endpoint.

## Conflicts resolved & regressions fixed

| Area | Issue if merged naively | Resolution |
|---|---|---|
| `UserController` OTP | PR's new helper called `getOtpCredential(getUsername())` **without** dev's `usernameTranscoder.encodeUsername(...)` (Keycloak-26 base32 decode, `ac04d637`) → 2FA reads as **disabled** for every OTP user | Keep the resilient helper, restore `encodeUsername(...)` inside it |
| `SecurityConfig` | PR added `GET /useradmin/tenantadmins` (no-suffix) → `hasAnyAuthority(TENANT_ADMIN, USER_ADMIN)`, exposing the cross-tenant admin **listing** to USER_ADMIN and breaking the existing `getTenantAdmins…returnForbidden` test | Keep only the intended `…/search` widening; drop the no-suffix matcher (falls through to the existing `…/**` → `TENANT_ADMIN` rule) |
| `TenantAdminUserService` | PR's `resolveTenantName()` swallowed **all** errors to `""`, masking real tenant-service outages | Keep dev's `tenantName()` (404 → skip, real errors propagate); drop the PR helper + its test asserting the masking behaviour |
| `UserControllerIT` | PR re-declared `@MockBean MagicLinkLoginService` / `EventNotificationService` that dev already has → **duplicate-field compile error** | Take dev's side (no duplicates) |
| `UserControllerGetUserDataTest` (new) | Missing `@Mock UsernameTranscoder` → `@InjectMocks` NPE on dev's `@NonNull` field | Add the mock; stub `encodeUsername` so the OTP-failure path is exercised |

`KeycloakUserDataProvider` fallback (token-based data when Keycloak is unreachable) is kept as-is; firstName/lastName/email are necessarily absent in that degraded path (not available from the token), which is acceptable for a Keycloak-down fallback.

## Open question for review

I kept dev's `tenantName()` over the PR's broader resilience. If you'd rather the tenant-admin search degrade gracefully on a tenant-service outage (instead of propagating), that's a small follow-up — but it should log, not silently blank, so it doesn't hide outages.

## Verification

`mvn test-compile` (JDK 17) + spotless clean; affected unit tests green (`UserControllerGetUserDataTest`, `KeycloakUserDataProviderTest`) — `Tests run: 10, Failures: 0, Errors: 0`. Full `*IT` suite not run locally (needs the `testing` H2 profile; CI covers compile/package).

Closes #65

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserController.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserController.java
@@ -80,6 +80,7 @@ import de.caritas.cob.userservice.api.helper.AuthenticatedUser;
 import de.caritas.cob.userservice.api.model.Chat;
 import de.caritas.cob.userservice.api.model.Consultant;
 import de.caritas.cob.userservice.api.model.EnquiryData;
+import de.caritas.cob.userservice.api.model.OtpInfoDTO;
 import de.caritas.cob.userservice.api.model.Session.SessionStatus;
 import de.caritas.cob.userservice.api.model.User;
 import de.caritas.cob.userservice.api.port.in.AccountManaging;
@@ -625,10 +626,7 @@ public class UserController implements UsersApi {
       var user = userAccountProvider.retrieveValidatedUser();
       partialUserData = askerDataProvider.retrieveData(user);
     }
-    var otpInfoDTO =
-        identityClientConfig.isOtpAllowed(authenticatedUser.getRoles())
-            ? identityManager.getOtpCredential(authenticatedUser.getUsername())
-            : null;
+    var otpInfoDTO = retrieveOtpCredentialIfAllowed();
 
     var fullUserData =
         userDtoMapper.userDataOf(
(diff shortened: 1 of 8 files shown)


Storypapst:
Verified this branch against the three correctness criteria for the #65 replacement; all hold and no code changes were needed:

1. **OTP path keeps `usernameTranscoder.encodeUsername(...)`** — the resilient `retrieveOtpCredentialIfAllowed()` helper still encodes the username before `getOtpCredential(...)`, so 2FA state is read correctly (no regression to the Keycloak-26 base32 path).
2. **SecurityConfig is not over-broadened** — only `GET /useradmin/tenantadmins/search` is widened to `hasAnyAuthority(TENANT_ADMIN, USER_ADMIN)`, placed before the `/useradmin/tenantadmins/**` matcher so it wins. The no-suffix cross-tenant listing still falls through to `TENANT_ADMIN` only.
3. **Tenant-name resolution surfaces real outages** — dev's `tenantName()` is retained: 404 → null (skip), any other `HttpClientErrorException` rethrows (no silent blanking).

Ran the affected unit tests locally on JDK 17: `UserControllerGetUserDataTest` + `KeycloakUserDataProviderTest` → Tests run: 5, Failures: 0, Errors: 0, BUILD SUCCESS. (The stack traces in test logs are the intentionally-thrown OTP/Keycloak exceptions being logged by the fallback paths, not failures.)

The Seer note about `firstName`/`lastName`/`email` being null in the Keycloak-down fallback is an accepted tradeoff for the degraded path (those fields aren't in the token) and is already documented in the PR body.

LGTM, merge-ready.

@coderabbitai review

PR #196, opened 2026-06-25
fix: persist Keycloak email, guard duplicate-account NPE, repair Matrix sync URL (#193)

### LINKED ISSUE #193: KeycloakService: updateEmail is a silent no-op + unguarded toLowerCase NPE (found via #126)
Two production issues surfaced while modernizing the JDK17 unit suite in #126. The tests were made to assert current behavior (with `TODO(FLAG)` comments) rather than green-wash these — they need a human decision because both look like deliberate Keycloak/Matrix-migration bypasses.

### 1. `updateEmail` is a no-op — email changes silently dropped in Keycloak
`src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakService.java:774-785`
The method logs `"Bypassing Keycloak email update for user..."` and returns; the real `userResource.update(representation)` (after `representation.setEmail(...)`) is commented out. It is reached via `changeEmailAddress`, `deleteEmailAddress` → `updateDummyEmail(userId)`, and `updateDummyEmail(String)`.
**Impact:** a user's email change is never persisted to Keycloak (login/notification email stays stale). If this bypass is no longer needed, restore the update; if intentional, document why and track removal.

### 2. Unguarded `.toLowerCase()` on possibly-null config → 409 becomes 500
`src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakService.java:417,423`
`handleCreateKeycloakUserError` calls `identityClientConfig.getErrorMessageDuplicatedEmail().toLowerCase()` and `getErrorMessageDuplicatedUsername().toLowerCase()` with no null guard. If either config value is unset at runtime, the duplicate-account path throws NPE → an opaque 500 instead of the intended 409 CONFLICT (EMAIL/USERNAME_NOT_AVAILABLE).
**Suggested fix:** null-guard (or `@NonNull`-validate) the two config getters.

_Discovered during the #126 `testFailureIgnore` removal; no production code was changed in that PR._

Storypapst:
Fixes the three production bugs tracked in #193 (surfaced while removing `testFailureIgnore` in #126). Each was pinned to its current broken behavior with a `TODO(FLAG)` / `TODO(PRODUCTION BUG)` marker; this PR applies the real fix and flips the tests to assert it.

## 1. `updateEmail` was a silent no-op — email changes never reached Keycloak
`KeycloakService#updateEmail` logged *"Bypassing Keycloak email update for testing"* and returned, with the real write commented out. Both self-service flows — `PUT /users/email` (change) and `DELETE /users/email` (reset to dummy) — updated the app DB, Rocket.Chat, the appointment service and fired a "your email changed" notification, while **Keycloak (the login + password-reset/notification authority) kept the old address**. 2FA-by-email and registration use the 2-arg overloads and were unaffected.

**Fix:** restore the real write **and** the `verifyEmail` duplicate-check, so a collision maps to `409 EMAIL_NOT_AVAILABLE` exactly like `updateUserData`. Dropped the `info` log (it printed the raw email — PII).

## 2. Unguarded `.toLowerCase()` on possibly-null config → `409` became an opaque `500`
`handleCreateKeycloakUserError` lower-cased the configured duplicate-email/-username markers with no null guard, so an unset marker NPE'd the duplicate-account path into a 500 (and a *blank* marker matched every error). 

**Fix:** null/blank-guard both markers; keep the `status == CONFLICT && contains(...)` fallback so duplicates still yield `409`.

**Bonus (same method):** `keycloakError` is an `@Value`-injected singleton field that was mutated as a return channel — a concurrency clobber. It now returns a local message instead.

## 3. Matrix `/sync` long-poll was dead — no messages ever delivered
`MatrixUrlBuilder` appended query values literally, so `syncRoom`'s JSON `filter` (`{"room":{...}}`) was read as a URI template and `buildAndExpand` threw `IllegalArgumentException: Map has no value for '"room"'`. `syncRoom` swallowed it, so `GET /sessions/{id}/sync` returned an empty result every time. `getRoomMessages` passes brace-free values and was unaffected.

**Fix:** bind query values as `q__`-prefixed URI variables (expand-then-encode keeps `!`/`:` literal while percent-encoding the JSON braces/quotes). Verified output: `…/sync?timeout=30000&filter=%7B%22room%22…%22!room:example.org%22…`. Also narrowed `syncRoom`'s `catch(Exception)` into `HttpStatusCodeException` (graceful) + `RuntimeException` (logs full stack) so a future URL/parse bug is never hidden as "no messages forever" again.

## Tests
- Flipped the three pinned tests (`changeEmailAddress…`, `deleteEmailAddress…`, `updateDummyMail_id…`, `syncRoomShouldUseDedicatedLongPollRestTemplate`) to assert the fixed behavior.
- Added regression tests: the null-marker NPE path (generic 500 + `409`-via-status fallback) and `syncRoom` returning parsed messages + resuming from the cached `next_batch` token.
- **Full unit suite: 1540 run, 0 failures, 0 errors, 7 skipped** (JDK 17).

Closes #193

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakService.java
@@ -6,6 +6,7 @@ import static de.caritas.cob.userservice.api.exception.httpresponses.customheade
 import static java.lang.Boolean.TRUE;
 import static java.util.Objects.isNull;
 import static java.util.Objects.nonNull;
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 import com.google.common.collect.Lists;
 import de.caritas.cob.userservice.api.adapters.keycloak.dto.KeycloakCreateUserResponseDTO;
@@ -375,6 +376,7 @@ public class KeycloakService implements IdentityClient {
     var locale =
         isNull(user.getPreferredLanguage()) ? "de" : user.getPreferredLanguage().toString();
     var kcUser = getUserRepresentation(user, firstName, lastName, locale);
+    String creationError = keycloakError;
     try (var response = keycloakClient.getUsersResource().create(kcUser)) {
       if (response.getStatus() == HttpStatus.CREATED.value()) {
         final String createdUserId = getCreatedUserId(response.getLocation());
@@ -394,14 +396,14 @@ public class KeycloakService implements IdentityClient {
         }
         return new KeycloakCreateUserResponseDTO(createdUserId);
       }
-      handleCreateKeycloakUserError(response);
+      creationError = handleCreateKeycloakUserError(response);
     }
     throw new InternalServerErrorException(
         String.format(
-            "Could not create Keycloak account for: %s %nKeycloak error: %s", user, keycloakError));
+            "Could not create Keycloak account for: %s %nKeycloak error: %s", user, creationError));
   }
 
-  private void handleCreateKeycloakUserError(Response response) {
+  private String handleCreateKeycloakUserError(Response response) {
     final int status = response.getStatus();
     String rawResponse = "";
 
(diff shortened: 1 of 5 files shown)


Storypapst:
Superseded by #194 (merged 2026-06-25, commit a0f2c695), which carries the identical KeycloakService email-persist fix, the duplicate-account NPE guard, and the MatrixUrlBuilder query-encoding repair — plus the accompanying unit tests. Closing this stale branch to avoid a divergent duplicate. No content is lost: everything here is already on `dev`.

PR #197, opened 2026-06-25
fix: avoid lazy registration profile failures

Storypapst:
## Summary
- load asker sessions and user-agency assignments through services while building user profile data
- keep agency language lookup inside a read-only transaction so consultant languages initialize for public registration requests
- add regression coverage for detached user proxies and transactional language lookup

## Tests
- JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home ./mvnw -Dtest=AskerDataProviderTest,ConsultantAgencyServiceTest test
- JAVA_HOME=/opt/homebrew/opt/openjdk@17/libexec/openjdk.jdk/Contents/Home ./mvnw test


Storypapst:
Tracked in OpenResilienceInitiative/ORISO-Frontend#280

Issue #223, opened 2026-06-29
[UserService] add unit tests for MagicLinkLoginService


Kadyan25:
**Type:** Improvement

## Problem

`MagicLinkLoginService` had zero unit test coverage. SMTP settings resolution, token lifecycle, username encoding fallback, and null safety were unverified by CI.

## Fix

30 tests added covering: blank/null username early returns, account not found, magic link enabled/disabled for users and consultants, SMTP settings resolution (null response, disabled, missing fields, string-typed values, nested value map, CTS throws), token expiry cleanup, username encoding fallback, consultant fallback, `normalizeBaseUrl` trailing slash.

## Verification

`./mvnw test -Dtest=MagicLinkLoginServiceTest` -- 30 tests, 0 failures

Storypapst:
Closing as completed — backlog re-audit found this already resolved: PR #227 merged — MagicLinkLoginService unit tests.

PR #227, opened 2026-06-29
add unit tests for MagicLinkLoginService

Kadyan25:
﻿#### [Issue #223](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/223)

## Summary

Adds 37 unit tests for `MagicLinkLoginService` — previously had zero coverage. The class handles magic-link login flows: token issuance, exchange, SMTP dispatch, and eligibility checks.

## Bug fixed

**`isMagicLinkAllowedForAccount` — NPE when `emailDummySuffix` is null**

`String.endsWith(null)` throws `NullPointerException`. The `@Value` field has a default but if the property is explicitly set to empty in a misconfigured environment, Spring injects null and the method crashes. The NPE test confirmed the bug before the fix was applied.

Before: `&& !target.getEmail().endsWith(emailDummySuffix)`
After: `&& (emailDummySuffix == null || !target.getEmail().endsWith(emailDummySuffix))`

## What was tested

- **Token TTL and single-use enforcement**: token removed before exchange; restored if Keycloak exchange fails or returns null
- **Eligibility checks**: disabled flag null vs false, dummy email suffix filter, blank email guard
- **SMTP validation**: blank host or from-address short-circuits to `ACCEPTED` without sending
- **Consulting type resolution**: nested value map unwrapping, trailing slash stripped from URL, invalid port string handled
- **Account lookup fallback**: user found by encoded username; consultant fallback when no user found
- **Cleanup scheduler**: `cleanupExpiredTokens` removes expired entries and keeps active ones
- **CTS transient failures**: ConsultingTypeService throwing returns `ACCEPTED` (fail-open)
- **Boolean settings as string**: `"TRUE"` string parsed correctly
- **Null emailDummySuffix (bug catch)**: no NPE, email treated as real address

## Approach

Mocked `UserService`, `ConsultantService`, `KeycloakService`, `MailService`, `ConsultingTypeServiceApiControllerFactory`. `ReflectionTestUtils` used to inject `emailDummySuffix` and `tokenStore`. `@MockitoSettings(LENIENT)` applied. `@Cacheable` annotations inactive without Spring proxy — cache behavior not tested (out of scope per team approach). `consumeMagicLink` full happy path deferred — token restore and fail-open paths are more critical for safety.

## Test results

Tests run: 37, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

## Files changed

| File | Action |
|---|---|
| `src/main/java/de/caritas/cob/userservice/api/service/auth/MagicLinkLoginService.java` | Bug fix (null guard) |
| `src/test/java/de/caritas/cob/userservice/api/service/auth/MagicLinkLoginServiceTest.java` | Created (37 tests) |

## Checklist
- [x] Bug fixed (null emailDummySuffix NPE)
- [x] Business logic covered (TTL, eligibility, SMTP, dummy suffix)
- [x] Error handling covered (exchange failure, transient infra, blank input)
- [x] Edge cases covered (null suffix, boolean-as-string, nested value map, invalid port)
- [x] Concurrency covered (token restore on exchange failure)
- [x] All 37 tests pass

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1203,6 +1203,33 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.jacoco</groupId>
+				<artifactId>jacoco-maven-plugin</artifactId>
+				<version>0.8.10</version>
+				<configuration>
+					<excludes>
+						<exclude>**/*generated*/**</exclude>
+						<exclude>**/org/openapitools/**</exclude>
+						<exclude>**/*$*</exclude>
+					</excludes>
+				</configuration>
+				<executions>
+					<execution>
+						<id>prepare-agent</id>
+						<goals>
+							<goal>prepare-agent</goal>
+						</goals>
+					</execution>
+					<execution>
+						<id>report</id>
+						<phase>test</phase>
+						<goals>
+							<goal>report</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
 		</plugins>
 	</build>
 	<profiles>
--- a/src/main/java/de/caritas/cob/userservice/api/service/auth/MagicLinkLoginService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/auth/MagicLinkLoginService.java
@@ -159,7 +159,7 @@ public class MagicLinkLoginService {
   private boolean isMagicLinkAllowedForAccount(AccountLoginTarget target) {
     return Boolean.TRUE.equals(target.getMagicLinkLoginEnabled())
         && isNotBlank(target.getEmail())
-        && !target.getEmail().endsWith(emailDummySuffix)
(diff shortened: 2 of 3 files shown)


shazia-k (commented):
Uncovered Code (~45%) That Was Not Tested
sendMagicLinkEmailSafely (45 lines) — never entered
buildHtml, generateAndStoreToken, buildMagicFrontendUrl — never called
consumeMagicLink happy path (line 119) — never returned a KeycloakLoginResponseDTO
exchangeTokenForUser catch block — only hit when admin succeeds but exchange throws
loginAdminForToken token-from-body path — success path never reached
resolveGlobalSmtpSettings returning a valid GlobalSmtpSettings — never reached
asIntSettingValue valid string "587" — only the invalid case was tested
resolveAccount decoded username fallback (lines 128–130) — never triggered
SMTP username / password blank guard — not tested

Please cover these scenarios

Issue #224, opened 2026-06-29
[UserService] add unit tests for SupervisorAddedEmailNotificationService


Kadyan25:
**Type:** Improvement

## Problem

`SupervisorAddedEmailNotificationService` had zero unit test coverage. Tenant ID resolution, template supplier URL path, localization branches, and null safety were unverified by CI.

## Fix

34 tests added covering: all three notify methods early returns, tenant ID resolution priority, template supplier URL resolution (happy path, throws, no url key), English/German localization branches for all 4 localize methods, `isLocalUrl` edge cases (127.0.0.1, ::1, malformed URL), hex color validation, HTML escaping of special chars, null sessionId in `buildSessionUrl`. Also added 5 tests for `SystemNotificationEmailSettingsService`.

## Verification

`./mvnw test -Dtest=SupervisorAddedEmailNotificationServiceTest,SystemNotificationEmailSettingsServiceTest` -- 39 tests, 0 failures

Storypapst:
Closing as completed — backlog re-audit found this already resolved: PR #228 merged — SupervisorAddedEmailNotificationService unit tests.

PR #228, opened 2026-06-29
add unit tests for SupervisorAddedEmailNotificationService

Kadyan25:
﻿#### [Issue #224](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/224)

## Summary

Adds 38 unit tests for `SupervisorAddedEmailNotificationService` — previously had zero coverage. The class handles email notifications for supervisor-added and email-address-changed events, including localization, tenant routing, SMTP dispatch, and URL building.

## Bug fixed

**`hasValidUserEmail` / `hasValidConsultantEmail` — NPE when `emailDummySuffix` is null**

Both methods called `String.endsWith(null)` which throws `NullPointerException`. NPE tests for both confirmed the defect before the fix was applied.

Before: `&& !user.getEmail().endsWith(emailDummySuffix)`
After: `&& (emailDummySuffix == null || !user.getEmail().endsWith(emailDummySuffix))`

## What was tested

- **German localization path**: when user language code is `de`, German-language template keys are resolved correctly
- **`notifyEmailAddressChanged` with non-null tenantData**: tenant-specific sender and reply-to fields populated from tenantData
- **UserService returning empty on lookup**: `getUser()` returning `Optional.empty()` handled gracefully, no NPE
- **SSL SMTP path**: when `sslEnabled` is true, SSL-specific SMTP configuration is applied
- **`buildSessionUrl` trailing slash**: base URL with trailing slash does not produce double slashes in the final link
- **Null emailDummySuffix (bug catch x2)**: `hasValidUserEmail` and `hasValidConsultantEmail` both return without NPE
- **Early returns**: missing SMTP settings, null tenant ID, blank email/username all short-circuit cleanly
- **URL fallback**: loopback/malformed `appBaseUrl` falls back to public frontend URL
- **Hex color validation**: valid hex accepted, invalid hex falls back to default
- **HTML escaping**: display names containing HTML characters are escaped before email template render

## Approach

Mocked `UserService`, `ConsultantService`, `MailService`, `TenantService`. `ReflectionTestUtils` used to inject `emailDummySuffix`. `@MockitoSettings(LENIENT)` applied.

## Test results

Tests run: 38, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

## Files changed

| File | Action |
|---|---|
| `src/main/java/de/caritas/cob/userservice/api/service/notification/SupervisorAddedEmailNotificationService.java` | Bug fix (null guard x2) |
| `src/test/java/de/caritas/cob/userservice/api/service/notification/SupervisorAddedEmailNotificationServiceTest.java` | Created (38 tests) |

## Checklist
- [x] Bug fixed (null emailDummySuffix NPE in hasValidUserEmail and hasValidConsultantEmail)
- [x] Business logic covered (localization, tenant data, session URL building)
- [x] Error handling covered (user not found fallback, early returns)
- [x] Edge cases covered (SSL SMTP, trailing slash, null suffix, hex validation, HTML escaping)
- [x] All 38 tests pass

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/service/notification/SupervisorAddedEmailNotificationService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/notification/SupervisorAddedEmailNotificationService.java
@@ -195,13 +195,13 @@ public class SupervisorAddedEmailNotificationService {
   private boolean hasValidUserEmail(User user) {
     return user != null
         && isNotBlank(user.getEmail())
-        && !user.getEmail().endsWith(emailDummySuffix);
+        && (emailDummySuffix == null || !user.getEmail().endsWith(emailDummySuffix));
   }
 
   private boolean hasValidConsultantEmail(Consultant consultant) {
     return consultant != null
         && isNotBlank(consultant.getEmail())
-        && !consultant.getEmail().endsWith(emailDummySuffix);
+        && (emailDummySuffix == null || !consultant.getEmail().endsWith(emailDummySuffix));
   }
 
   private void sendEmailSafely(
--- a/src/test/java/de/caritas/cob/userservice/api/service/notification/SupervisorAddedEmailNotificationServiceTest.java
+++ b/src/test/java/de/caritas/cob/userservice/api/service/notification/SupervisorAddedEmailNotificationServiceTest.java
@@ -549,6 +549,92 @@ class SupervisorAddedEmailNotificationServiceTest {
         .doesNotThrowAnyException();
   }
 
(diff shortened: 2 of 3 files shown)


shazia-k (commented):
localizeSupervisorAssignmentRemoved German path — no test reaches this with a German-language user
notifyEmailAddressChanged + non-null tenantData — always called with null tenantData in tests
userService.getUser() returning Optional.empty() — fallback-to-original-user path untested
SSL SMTP path (isSecure() == true) — all settings use secure = false
buildSessionUrl trailing-slash strip — no base URL ending in / is ever tested

Need some more improvements

Issue #234, opened 2026-06-30
add unit tests for AdminAgencyRelationService


Kadyan25:
## What
Add unit tests for \AdminAgencyRelationService\ to bring coverage above 80%.

## Why
Class had 0% test coverage. Covers all 4 public methods: create, delete, synchronize, and appendAgenciesForAdmins including empty-set and multi-admin scenarios.

## Coverage
- Instruction: 86%
- Branch: 60% (structural — @SneakyThrows generated bytecode)
- Tests: 10

Storypapst:
Closing as completed — backlog re-audit found this already resolved: PR #235 merged — AdminAgencyRelationService unit tests.

PR #235, opened 2026-06-30
add unit tests for AdminAgencyRelationService

Kadyan25:
﻿#### [Issue #234](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/234)

## Summary

Adds 10 unit tests for `AdminAgencyRelationService` — previously had zero coverage. The class orchestrates admin-agency relationship CRUD and enrichment operations via three delegate services.

## What was tested

- **`createAdminAgencyRelation` — happy path**: delegates to `CreateAdminAgencyRelationService.create()` with correct adminId and DTO, verified with exact argument matching
- **`deleteAdminAgencyRelation` — happy path**: repo lookup succeeds, entity deleted
- **`deleteAdminAgencyRelation` — not found**: repo returns empty → `CustomValidationHttpStatusException` thrown, delete never called
- **`synchronizeAdminAgenciesRelation` — happy path**: delegates to `SynchronizeAdminAgencyRelation.synchronizeAdminAgenciesRelation()` with correct adminId and agency list, verified with exact argument matching
- **`appendAgenciesForAdmins` — empty admin set**: empty set passed in, method returns early without processing
- **`appendAgenciesForAdmins` — single admin with matching agency**: agency resolved and populated on DTO; both `getId()` and `getName()` asserted to verify `BeanUtils.copyProperties` copies fields correctly
- **`appendAgenciesForAdmins` — admin with no matching agencies (different adminId)**: empty agency list set on DTO
- **`appendAgenciesForAdmins` — no relations exist**: empty list from repo, empty agencies set on DTO
- **`appendAgenciesForAdmins` — multiple admins**: each admin resolves its own agency list independently
- **`appendAgenciesForAdmins` — agencyId not in returned agencies**: relation exists but agency service doesn't return a matching agency — empty list set on DTO

## Approach

Mocked `CreateAdminAgencyRelationService`, `AdminAgencyRepository`, `SynchronizeAdminAgencyRelation`, `AgencyAdminService`. `AdminAgencyBase` is a Spring Data projection interface — instantiated via `mock(AdminAgencyBase.class)`. `BeanUtils.copyProperties` in `buildCopiedAgency` generates an implicit try-catch in bytecode via `@SneakyThrows` — the catch path cannot be triggered without BeanUtils actually throwing, so branch coverage is 60% (structural, not a gap in business logic coverage). `@MockitoSettings(LENIENT)` applied.

## Test results

Tests run: 10, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

## Files changed

| File | Action |
|---|---|
| `src/test/java/de/caritas/cob/userservice/api/admin/service/admin/AdminAgencyRelationServiceTest.java` | Created |

No production code changes.

## Checklist
- [x] All public methods covered (create, delete, sync, append)
- [x] Happy paths covered
- [x] Error handling covered (not-found throws exception, delete never called)
- [x] Edge cases covered (empty admin set, no matching agency, agencyId mismatch, multiple admins)
- [x] Field copy correctness verified (getName() asserted alongside getId())
- [x] Delegation verified with exact argument matching via verify()
- [x] No production code changes
- [x] All 10 tests pass

### DIFF
--- a/src/test/java/de/caritas/cob/userservice/api/admin/service/admin/AdminAgencyRelationServiceTest.java
+++ b/src/test/java/de/caritas/cob/userservice/api/admin/service/admin/AdminAgencyRelationServiceTest.java
@@ -115,6 +115,7 @@ class AdminAgencyRelationServiceTest {
 
     assertThat(admin.getAgencies()).hasSize(1);
     assertThat(admin.getAgencies().get(0).getId()).isEqualTo(10L);
+    assertThat(admin.getAgencies().get(0).getName()).isEqualTo("Test Agency");
   }
 
   @Test
--- a/src/test/java/de/caritas/cob/userservice/api/admin/service/admin/AdminAgencyRelationServiceTest.java
+++ b/src/test/java/de/caritas/cob/userservice/api/admin/service/admin/AdminAgencyRelationServiceTest.java
@@ -1,0 +1,206 @@
+package de.caritas.cob.userservice.api.admin.service.admin;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import de.caritas.cob.userservice.agencyadminserivce.generated.web.model.AgencyAdminResponseDTO;
+import de.caritas.cob.userservice.api.adapters.web.dto.AdminDTO;
+import de.caritas.cob.userservice.api.adapters.web.dto.CreateAdminAgencyRelationDTO;
+import de.caritas.cob.userservice.api.admin.service.admin.create.agencyrelation.CreateAdminAgencyRelationService;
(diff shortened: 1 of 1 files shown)


Shirloin (dismissed):
The PR description doesn't match the actual test file:

1. "null admin set: null input handled without NPE". This test doesn't exists in the file, and production code (admins.stream()) would actually throw NPE on null input since there's no null check. Please either add this test or remove the claim.
2. "ArgumentCaptor on createAdminAgencyRelation" and "ArgumentCaptor on synchronizeAdminAgenciesRelation" — neither test uses ArgumentCaptor, both just verify() with plain equality matchers. Please update the description to match, or add the captor verification if that's the intent.

One more thing worth checking: buildCopiedAgency() calls BeanUtils.copyProperties(result, agency) — Spring's BeanUtils typically expects (target, source), so this looks correct, but Apache Commons BeanUtils (which this import appears to be, given @SneakyThrows is needed) expects (dest, orig) too. Worth double-checking this actually copies fields correctly rather than silently copying nothing — none of the appendAgenciesForAdmins tests assert on agency.getName() or other copied fields beyond getId(), which wouldn't catch a reversed-copy bug.

Kadyan25:
fixed it @Shirloin

Shirloin (approved):

Issue #251, opened 2026-07-01
fix flaky ArchiveOrDeleteSessionStatisticsEventTest — LocalDateTime.now() drops :00 seconds in toString()


Kadyan25:
﻿## What
Fix flaky unit test in `ArchiveOrDeleteSessionStatisticsEventTest` — `getPayload_Should_ReturnValidJsonPayload` fails non-deterministically on CI.

## Why
`LocalDateTime.now()` is used as a static field initializer. The assertion then calls `.truncatedTo(ChronoUnit.SECONDS).toString() + "Z"` to build the expected value. `LocalDateTime.toString()` silently drops trailing zero seconds — so when `now()` lands on `:00` seconds (e.g. `10:54:00`), `toString()` produces `"2026-06-30T10:54"` instead of `"2026-06-30T10:54:00"`. The serializer in `getPayload()` always emits the full `":00Z"` suffix, so the strings never match at that second. This causes intermittent CI failures on any branch that inherits this test from `dev`.

## Fix
Replace `LocalDateTime.now()` with a fixed timestamp that never ends in `:00` seconds:
```java
private static final LocalDateTime END_DATE = LocalDateTime.of(2024, 1, 15, 10, 30, 45);
```

Storypapst:
My coding agent checked the current issue/PR record. Technical follow-up: @shazia-k. [PR #252](https://github.com/OpenResilienceInitiative/ORISO-UserService/pull/252) is closed. The targeted test-fix PR #252 is closed unmerged; the consent-pointer cross-reference is unrelated delivery. Link a current fixed-timestamp implementation or successor.

Storypapst:
Technical acceptance audit (code/review lane, 2026-09-09)

The issue has no checklist. Original PR #252 closed unmerged, but successor PR #468 is contained in `origin/dev`; `ArchiveOrDeleteSessionStatisticsEventTest` now uses a fixed non-zero-seconds timestamp and the canonical ISO formatter. The flake cause is technically removed. Link #468 as the delivery record; normal CI execution remains the final proof.

PR #241, opened 2026-06-30
add unit tests for SupervisorLogsService

Kadyan25:
﻿#### [Issue #240](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/240)

## Summary

Adds 16 unit tests for `SupervisorLogsService` — previously had zero coverage. Nearly identical structure to `InactiveAccountAuditLogsService`: same JWT tenant resolution, same pagination clamping, same JDBC pattern but over a UNION ALL query for supervisor add/remove events.

## What was tested

- **Happy path**: valid TenantContext, page 1, 10 per page — returns correct total and data list
- **Multiple entries**: result set with 4 entries mapped correctly
- **Null JDBC return**: `queryForObject` returning null defaults total to 0, no NPE
- **Pagination clamping — perPage too large**: 9999 clamped to 200
- **Pagination clamping — perPage zero**: 0 clamped to 1
- **Pagination clamping — negative page**: -3 clamped to 1
- **JWT numeric tenantId**: parsed and used for JDBC query
- **JWT string tenantId**: `"3"` string parsed to Long correctly
- **JWT tenantId = 0** (super-admin context): query runs without tenant filter
- **Fallback — no token**: null access token uses `TenantContext.getCurrentTenant()`
- **Fallback — blank token**: blank string triggers TenantContext fallback
- **Fallback — malformed JWT**: TenantContext fallback, no exception
- **Fallback — malformed JSON payload**: fallback gracefully, no exception
- **Fallback — blank tenantId string**: `"tenantId": ""` falls back to TenantContext
- **Paged result**: page 3 of 100 total — page number preserved
- **tenantId bound into SqlParams (JWT path)**: `ArgumentCaptor<SqlParameterSource>` verifies `tenantId=55L` is actually bound
- **tenantId bound into SqlParams (TenantContext path)**: captures `tenantId=77L` from TenantContext

## Approach

Mocked `NamedParameterJdbcTemplate` and `AuthenticatedUser`. `TenantContext.setCurrentTenant()` called directly. `TenantContext.clear()` in `@AfterEach` prevents test bleed. `ArgumentCaptor<SqlParameterSource>` verifies the param key `"tenantId"` is bound — not just that JDBC was called. `@MockitoSettings(LENIENT)` applied.

## Test results

Tests run: 16, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

## Files changed

| File | Action |
|---|---|
[...]

PR #237, opened 2026-06-30
add unit tests for SessionStatisticsResultDTO

Kadyan25:
﻿#### [Issue #236](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/236)

## Summary

Adds 20 unit tests for `SessionStatisticsResultDTO` — the generated OpenAPI DTO for session statistics, previously with zero dedicated tests. Full 100% instruction and branch coverage achieved. Also tightens a loose toString assertion and removes a logically incorrect hashCode test.

## What was tested

- **Fluent setters**: all 8 setter methods (`id`, `rcGroupId`, `consultingType`, `agencyId`, `isTeamSession`, `postcode`, `messageDate`, `createDate`) return `this` for chaining
- **Getter/setter round-trip**: each field stored and retrieved correctly
- **`equals` — reflexive**: object equals itself
- **`equals` — null**: object does not equal null
- **`equals` — wrong class**: object does not equal a String
- **`equals` — all fields equal**: two fully-populated instances with same values are equal
- **`equals` — per-field inequality**: one test per field (8 tests) — changing any single field breaks equality
- **`hashCode` — consistent**: same object returns same hash on repeated calls
- **`hashCode` — equal objects same hash**: two equal instances produce the same hashCode
- **`toString`**: output contains all field names (`id`, `rcGroupId`, `consultingType`, `agencyId`, `isTeamSession`, `postcode`, `messageDate`, `createDate`)

## Removed test

`hashCode_Should_DifferForDifferentObjects` was removed — Java's hashCode contract does NOT require unequal objects to have different hash codes (hash collisions are allowed). The assertion `isNotEqualTo` on hashCodes of unequal objects was logically incorrect and would produce false failures if Lombok's generated hash happened to collide.

## Approach

No mocks needed — pure logic test on a generated DTO. The class has no dependencies. `@MockitoSettings(LENIENT)` not needed. toString assertion uses a field-name-specific string (`"consultingType"`) rather than a bare value (`"1"`) to avoid matching incidental matches in date fields.

## Test results

Tests run: 20, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

## Files changed

| File | Action |
|---|---|
| `src/test/java/de/caritas/cob/userservice/api/statistics/model/SessionStatisticsResultDTOTest.java` | Created |

No production code changes.

## Checklist
- [x] All fluent setters covered
- [x] equals contract covered (reflexive, null, wrong class, equal, per-field inequality)
[...]

PR #252, opened 2026-07-01
fix flaky ArchiveOrDeleteSessionStatisticsEventTest — replace LocalDateTime.now() with fixed timestamp

Kadyan25:
﻿#### [Issue #251](https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/251)

## Summary

Fixes a flaky test in `ArchiveOrDeleteSessionStatisticsEventTest` that caused non-deterministic CI failures on any branch inheriting this test from `dev`. One-line change — no production code touched.

## Bug fixed

**`getPayload_Should_ReturnValidJsonPayload` — fails when `LocalDateTime.now()` lands on `:00` seconds**

`LocalDateTime.toString()` silently drops trailing zero seconds. When `now()` is e.g. `10:54:00`, `.truncatedTo(ChronoUnit.SECONDS).toString()` produces `"2026-06-30T10:54"` instead of `"2026-06-30T10:54:00"`. The serializer in `getPayload()` always emits the full `":00Z"` suffix, so the strings don't match at that exact second — test fails.

Before: `private static final LocalDateTime END_DATE = LocalDateTime.now();`
After: `private static final LocalDateTime END_DATE = LocalDateTime.of(2024, 1, 15, 10, 30, 45);`

A fixed timestamp with non-zero seconds eliminates the race condition entirely.

## What was tested

- **`getEventType_Should_ReturnEventTypeArchiveMessage`**: still passes — no change to event type logic
- **`getPayload_Should_ReturnValidJsonPayload`**: now deterministic — fixed timestamp always serializes to `"2024-01-15T10:30:45Z"`, assertion always matches

## Approach

No mocks needed. Single static field replacement. `LocalDateTime.of(2024, 1, 15, 10, 30, 45)` was chosen as a clearly artificial fixed date with non-zero seconds that can never produce the `:00` edge case. `spotless:apply` run before commit — no formatting changes needed.

## Impact

Once merged to `dev`, the Feature Branch CI on PR #237 and PR #241 will stop failing on this test. Both PRs are currently blocked from approval due to this flaky failure.

## Test results

Tests run: 2, Failures: 0, Errors: 0, Skipped: 0 — BUILD SUCCESS

## Files changed

| File | Action |
|---|---|
| `src/test/java/de/caritas/cob/userservice/api/service/statistics/event/ArchiveOrDeleteSessionStatisticsEventTest.java` | 1 line changed |

No production code changes.

## Checklist
- [x] Bug fixed (flaky LocalDateTime.now() replaced with fixed timestamp)
- [x] Both existing tests pass
- [x] No production code changes
- [x] Unblocks PR #237 and PR #241 once merged to dev

### DIFF
--- a/src/test/java/de/caritas/cob/userservice/api/service/statistics/event/ArchiveOrDeleteSessionStatisticsEventTest.java
+++ b/src/test/java/de/caritas/cob/userservice/api/service/statistics/event/ArchiveOrDeleteSessionStatisticsEventTest.java
@@ -22,7 +22,7 @@ class ArchiveOrDeleteSessionStatisticsEventTest {
   private ArchiveOrDeleteSessionStatisticsEvent archiveOrDeleteSessionStatisticsEvent;
   private static final Long TENANT_ID = 2L;
   private static final String USER_ID = "userId";
-  private static final LocalDateTime END_DATE = LocalDateTime.now();
+  private static final LocalDateTime END_DATE = LocalDateTime.of(2024, 1, 15, 10, 30, 45);
 
   @BeforeEach
   public void setup() throws NoSuchFieldException, IllegalAccessException {


Kadyan25:
Closing in favour of #250 which fixes the same flaky test with a better approach (asserting via CustomLocalDateTime.toIsoTime instead of a fixed timestamp).

PR #293, opened 2026-07-03
Move supervisor asides to a separate room the client is never in (ADR-008 safeguarding)

Storypapst:
## Summary

Closes the ADR-008 side-channel leak on the backend: supervisor feedback (`[SUPERVISOR_FEEDBACK]`) and coordinator↔peer asides (`[VISIBLE_TO:…]`) were posted into the **same Matrix room as the client** and hidden only by a React `return null`. With Megolm off, the plaintext aside is delivered to the client's own device and is trivially recoverable (DevTools `/sync`, any Matrix client). For U25 minors this is a safeguarding hazard, not a cosmetic bug.

This is the **backend half** — it provisions the separate room so asides can be routed out of the client's room. The paired frontend change (routing aside *sends* to this room + rendering it for authorized members) is a follow-up on the Matrix-only frontend branch.

## What changed

`SessionSupervisorFacade.addSupervisor` now:
- provisions (or reuses) a **per-session supervision side room** the client is **never** invited to. It is created with the `private_chat` preset → `join_rule: invite`, so the client cannot join even if the room id leaks. One side room is shared by all supervisors of a session.
- invites + joins the supervisor (and, if different, the assigned consultant) into that side room;
- **still** invites the supervisor to the client room read-only (power level 10) for observation — only their *feedback* moves out (exactly the ADR-008 decision);
- stores the **side-room id** in `SessionSupervisor.matrix_room_id`, **reusing the existing column** — no schema change (a new `@Column` would crash on boot: `ddl-auto=validate` is on everywhere, the ADR-006 hazard). The client room id is always available via `session.getMatrixRoomId()`.
- provisions the side room **before** touching the client room, so a failure leaves no half state.

`removeSupervisor` now removes the supervisor from **both** rooms (side + client). Old-style rows that stored the client room id in `matrix_room_id` are **never** reused as a side room (that would re-open the leak) — a fresh side room is created.

## Validation (verified live on Pre-Dev)

- New `SessionSupervisorFacadeTest` (4 tests): stores side-room id not client-room id; provisions the side room and invites into both rooms; reuses an existing side room; never treats an old-style client-room row as a side room. Green.
- Live proof on Pre-Dev (hot-deployed image): added a supervisor → a **new** side room `!Zrr…` distinct from the client room `!DbE…`; the side room contains only the two counsellors, **not the asker** (`join_rule: invite`). Sent a `[SUPERVISOR_FEEDBACK]` aside to the side room → the **asker's `/sync` does not contain it** (leak closed), while a side-room member's `/sync` does.

## Follow-ups (tracked, not in this PR)

- Frontend: route aside sends (`isSupervisor` / `VISIBLE_TO`) to the side room and render it for authorized members (pairs with ORISO-Frontend Matrix-only branch).
- ADR-008 AXIS 2 (separate): tighten `addSupervisor` authority (assigned-consultant-only) + record consent/justification.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/facade/SessionSupervisorFacade.java
+++ b/src/main/java/de/caritas/cob/userservice/api/facade/SessionSupervisorFacade.java
@@ -136,7 +136,20 @@ public class SessionSupervisorFacade {
       throw new InternalServerErrorException("Failed to create consultant Matrix token");
     }
 
-    // Invite supervisor to Matrix room
+    // ADR-008: provision (or reuse) a SEPARATE supervision side room that the client is NEVER
+    // invited to. Supervisor↔counsellor feedback and coordinator↔peer asides live here, so they
+    // are never delivered to the client's Matrix device. Do this BEFORE touching the client room
+    // so a failure leaves no half state. The supervisor still joins the client room read-only
+    // (below) for observation — only their feedback moves out.
+    String sideRoomId =
+        ensureSupervisionSideRoom(
+            session,
+            addedByConsultant,
+            supervisorConsultant,
+            supervisorMatrixUserId,
+            consultantToken);
+
+    // Invite supervisor to the CLIENT room as a read-only observer.
     try {
       matrixSynapseService.inviteUserToRoom(matrixRoomId, supervisorMatrixUserId, consultantToken);
       log.info(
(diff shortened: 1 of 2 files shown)


Storypapst:
Paired frontend PR: OpenResilienceInitiative/ORISO-Frontend#367 (routes aside sends to this side room + renders them for members). This backend PR provisions the room; #367 makes the running app use it. Review together.

PR #295, opened 2026-07-04
fix: shut down Matrix room when a group chat is stopped or deleted

Storypapst:
## What users experienced

When a consultant stopped a group chat, the chat kept looking alive for everyone in it. No message, no removal — askers and consultants could still open the room and try to write. The reason: the stop path still calls the old Rocket.Chat cleanup, which silently does nothing since the Matrix migration (`rocket-chat.enabled=false`). Only the database row was deleted; the Matrix room was never touched.

The frontend now shows a "group chat stopped" overlay when it sees a Matrix lifecycle signal (`Room.myMembership` becoming leave/ban, or `m.room.tombstone`) — added in OpenResilienceInitiative/ORISO-Frontend#359, where this gap was found during review. But no signal ever arrived from the backend.

## What this PR does

When a group chat is stopped (and when the last member leaves and the chat is deleted), the backend now shuts down the chat's Matrix room. All members are removed from the room, so their clients get the membership signal and the overlay from FE #359 triggers.

Covered paths:
- **Stop group chat** (`StopChatActionCommand`, used by the stop endpoint and the stale-chat deactivation workflow)
- **Last member leaves** (`JoinAndLeaveChatFacade.leaveChat`, which also deletes the chat)

## Design choice: room shutdown (kick-all) instead of tombstone

We reuse the existing Synapse admin room deletion call (`MatrixSynapseService#purgeRoom`, `DELETE /_synapse/admin/v2/rooms/{roomId}`). It removes every member from the room and purges it. The frontend listens for both kick and tombstone, so either works — but the shutdown call:

- already exists in this codebase (the account deletion workflow uses it), so there is one room teardown mechanism, not two
- needs no extra power-level or state-event plumbing
- leaves nothing behind: members are out and the room is gone

Behavior guards:
- Chats without a `matrix_room_id` (legacy) are skipped silently
- A failing Matrix call is logged but never fails the stop — the database deletion stays the source of truth
- The repetitive-chat re-creation path (`ChatReCreator`) is untouched

## Test evidence

TDD (red first, then green). New tests:
- stop with Matrix room id → Synapse room shutdown called, chat deleted
- stop with null room id → no Matrix call, no error
- Matrix returns failure or throws → stop still succeeds
- repetitive chat → no room shutdown
- last member leaves → chat deleted + room shut down
- shutdown service unit tests (null/blank guard, failure tolerance)

Full build: `./mvnw -B package` → **BUILD SUCCESS, 2191 tests, 0 failures, 0 errors** (no `testFailureIgnore` in the current pom — failures would fail the build). Spotless applied.

Found during review of OpenResilienceInitiative/ORISO-Frontend#359.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/actions/chat/ChatReCreator.java
+++ b/src/main/java/de/caritas/cob/userservice/api/actions/chat/ChatReCreator.java
@@ -1,31 +1,52 @@
 package de.caritas.cob.userservice.api.actions.chat;
 
 import static de.caritas.cob.userservice.api.helper.CustomLocalDateTime.nowInUtc;
-import static java.util.Objects.nonNull;
+import static java.util.Objects.isNull;
+import static org.apache.commons.lang3.StringUtils.isBlank;
 
-import de.caritas.cob.userservice.api.adapters.rocketchat.RocketChatService;
+import de.caritas.cob.userservice.api.adapters.matrix.MatrixSynapseService;
 import de.caritas.cob.userservice.api.exception.httpresponses.InternalServerErrorException;
-import de.caritas.cob.userservice.api.exception.rocketchat.RocketChatAddUserToGroupException;
-import de.caritas.cob.userservice.api.exception.rocketchat.RocketChatCreateGroupException;
-import de.caritas.cob.userservice.api.exception.rocketchat.RocketChatUserNotInitializedException;
-import de.caritas.cob.userservice.api.helper.RocketChatRoomNameGenerator;
+import de.caritas.cob.userservice.api.exception.matrix.MatrixCreateRoomException;
 import de.caritas.cob.userservice.api.model.Chat;
 import de.caritas.cob.userservice.api.service.ChatService;
+import java.time.Instant;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.springframework.stereotype.Service;
 
+/**
+ * Re-creates the messenger room of a repetitive group chat for its next occurrence (ADR-004:
+ * Matrix-native, Rocket.Chat is disabled).
+ *
+ * <p>The order of operations is fail-safe: the new Matrix room is created first, and only after
+ * that succeeded the old room is shut down (best-effort, via {@link MatrixChatShutdownService}) so
+ * its members receive the membership lifecycle signal. There is no lazy room provisioning for group
+ * chats ({@code StartChatFacade} only activates a chat whose room already exists), so a failed
+ * re-creation surfaces as an error while the chat stays untouched and usable — stopping the chat
+ * again retries the whole operation.
+ */
+@Slf4j
 @Service
 @RequiredArgsConstructor
 public class ChatReCreator {
 
-  private static final RocketChatRoomNameGenerator roomNameGenerator =
-      new RocketChatRoomNameGenerator();
+  private static final String ROOM_ALIAS_PREFIX = "group_chat_";
 
   private final ChatService chatService;
-  private final RocketChatService rocketChatService;
+  private final MatrixSynapseService matrixSynapseService;
+  private final MatrixChatShutdownService matrixChatShutdownService;
 
-  public void updateAsNextChat(Chat chat, String rcGroupId) {
-    chat.setGroupId(rcGroupId);
+  /**
+   * Resets the given chat to its next occurrence: the new Matrix room id is persisted both as
+   * legacy group id and as Matrix room id (mirroring chat creation) and the chat is deactivated
+   * until it is started again.
+   *
+   * @param chat the repetitive {@link Chat} to update
+   * @param matrixRoomId the id of the freshly created Matrix room
+   */
+  public void updateAsNextChat(Chat chat, String matrixRoomId) {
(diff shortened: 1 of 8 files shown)


shazia-k (commented):
The two breaking bugs (both in StopChatActionCommand)
Bug 1 — Repetitive chat stop always throws HTTP 500 chatReCreator.recreateMessengerChat calls rocketChatService.createPrivateGroupWithSystemUser, which DisabledRocketChatService implements by returning Optional.empty(). The orElseThrow in recreateMessengerChat fires a RocketChatCreateGroupException, which is caught and re-thrown as InternalServerErrorException. Every repetitive chat stop fails before the DB is touched, let alone Matrix.
Bug 2 — Repetitive chats never get a Matrix lifecycle signal matrixChatShutdownService.shutdownRoom(chat) is inside the else branch, so it is only reached for non-repetitive chats. The test stopChatShouldNotShutDownMatrixRoomWhenChatIsRepetitive explicitly asserts verifyNoInteractions(matrixSynapseService), documenting this confirmed gap. Members of a weekly group chat remain in a live-looking Matrix room after the occurrence ends — exactly the symptom described in the issue.

Storypapst:
@shazia-k Thank you for the sharp review — both findings were real and are fixed in f9aea6e6. Point by point:

### Bug 1 — repetitive chat stop always threw HTTP 500

Confirmed exactly as you described. I reproduced it first in a test that wires the real `DisabledRocketChatService` into `StopChatActionCommand` — before the fix it failed with `InternalServerErrorException: Error while creating private group in Rocket.Chat ...` (red), and it passes now (green).

The fix rewrites `ChatReCreator` to be Matrix-native. It no longer depends on `RocketChatService` at all: the next-occurrence room is created via `MatrixSynapseService#createRoomAsMatrixUser` as the chat owner's Matrix user — the same provisioning call `CreateChatFacade` uses when a group chat is created. The new room id is persisted on both columns (`rc_group_id` for legacy readers and `matrix_room_id`), mirroring chat creation.

### Bug 2 — repetitive chats never got the Matrix lifecycle signal

Also fixed. `ChatReCreator#recreateMessengerChat` now shuts down the old Matrix room via the existing `MatrixChatShutdownService` (same kick-all + purge mechanism as the non-repetitive path), so members of a weekly chat get the membership signal and the FE #359 overlay triggers. The test that documented the gap (`stopChatShouldNotShutDownMatrixRoomWhenChatIsRepetitive` with its `verifyNoInteractions(matrixSynapseService)`) is replaced by tests asserting the new behavior: old room purged, new room provisioned, chat row updated.

### Fail-safe semantics — why the re-creation fails loud

I checked for a lazy provisioning path and there is none: `StartChatFacade#startChat` only flips `active` and 500s when the group id is missing — nothing ever creates a room for an existing chat. Silently persisting a room-less next occurrence would therefore leave a permanently dead chat. So the re-creation fails loud, but in a safe order:

1. Create the new Matrix room **first**. If this fails → `InternalServerErrorException`, and nothing has been touched: the chat stays active, the old room stays alive, the row is unchanged — stopping again simply retries.
2. Only after the new room exists is the old room shut down — still best-effort: a failing shutdown is logged and never breaks the stop (worst case an orphaned room, same semantics as the non-repetitive path).
3. Persist the new room id and reset the chat state (`updateAsNextChat`).

Covered by `stopRepetitiveChatShouldFailAndLeaveChatUntouchedWhenMatrixRoomCreationFails` and the `ChatReCreatorTest` cases (including in-order verification that creation happens before the old-room shutdown).

### Other callers of the same code

- `JoinAndLeaveChatFacade#leaveChat` calls the same `chatReCreator.recreateMessengerChat` for repetitive chats when the last member leaves — it had the identical latent 500 and is fixed for free by the shared `ChatReCreator` rewrite (the facade itself needed no change).
- The stale-chat sweep (`DeactivateGroupChatService` → `StopChatActionCommand`) previously hit the same 500 on every repetitive chat it tried to deactivate; it now works. If Matrix is down during a sweep, that run aborts and the next cron tick retries — strictly better than before.
- `CreateChatFacade` still contains an RC `createPrivateGroupWithSystemUser` call, but it is only reachable when `rocket-chat.enabled=true` (both createChat V1/V2 route to the Matrix flow otherwise), so there is no live 500 there.

### Test evidence

New/updated tests (TDD, red first):
- `stopRepetitiveChatShouldRecreateMatrixRoomAndShutDownOldRoomWhenRocketChatIsDisabled` — real `DisabledRocketChatService`, real `ChatReCreator`: no exception, old room purged, new room created as chat owner, both id columns updated, start date +1 week, chat deactivated
- `stopRepetitiveChatShouldFailAndLeaveChatUntouchedWhenMatrixRoomCreationFails` — fail-loud path: old room untouched, nothing saved
- new `ChatReCreatorTest` (5 tests): create-as-owner + shutdown ordering, creation failure keeps old room alive, missing room id in the response, owner without Matrix user, `updateAsNextChat` state reset
- all non-repetitive path tests unchanged and green

Full build: `./mvnw -B package` → **BUILD SUCCESS**, 2171 tests, 0 failures, 0 errors (7 skipped), JDK 21. Spotless applied.

Shirloin (approved):

PR #297, opened 2026-07-04
fix: make group-chat member queries Matrix-native

Storypapst:
## What users saw before this fix

With Rocket.Chat switched off (the default since ADR-004), the app still asked Rocket.Chat "who is in this chat?" — and always got back an empty list. Four group-chat features quietly broke:

- **Empty member lists.** Opening a group chat's member list always showed nobody.
- **Removed consultants kept access.** Detaching an agency from a consultant was supposed to kick them out of that agency's team chats. It never did — they stayed in the room.
- **Ban was broken.** Banning an adviceseeker from a group chat could 500 or do nothing.
- **No live updates.** Live events for a group chat (new message pings, push notifications) reached no one.

## What this changes

All four now read chat membership from the **Matrix room** (the only chat backend that is actually running), reusing the `GroupChatMembershipService` from #296.

| Caller | Fix |
| --- | --- |
| `GetChatMembersFacade.getChatMembers` | Reads the chat's Matrix room members, maps each Matrix user back to its consultant / adviceseeker account (username + display name), returns the **same DTO shape** as before. |
| `RemoveConsultantFromRocketChatService.observeConsultantsToRemove` | Finds the consultants in the session's Matrix room that should lose access (every consultant except the assigned one) and removes them from the room. Askers and the assigned consultant are never touched. |
| `Messenger` ban path | `banUserFromChat` now performs a real **Matrix room ban** (new `MatrixRoomClient.banUserFromRoom` / `MatrixSynapseService.banUserFromRoomAsModerator`, acting as the chat owner). `isInChat` reads Matrix membership instead of `findMembers(...).orElseThrow()`, so it no longer throws. |
| `RelevantUserAccountIdsByChatProvider.getChatUsers` | Collects live-event recipients from the Matrix room membership. |

## Safety

- **One Synapse call per query**, then plain DB look-ups per member — no per-member Synapse loop on hot paths.
- **Technical accounts filtered** (Synapse admin, agency service accounts, group-chat system users) exactly as in #296.
- **Fail-safe direction respected.** Empty-on-uncertainty is only ever treated as a read-only "nobody to show / nobody to notify"; it never triggers a destructive action. `isInChat` fails safe to `false` (no wrongful removal), and consultant removal does nothing when the room state can't be read.
- `RocketChatRollbackService` stays inert (nothing to roll back in the Matrix world).

## Product decisions to confirm

1. **Ban = Matrix room ban.** The old Rocket.Chat behaviour was a "mute". A Matrix ban both removes the user *and* blocks re-join, which is stronger and cleaner. `banUserFromChat` returns `false` (→ `404 user not found in chat`) when it genuinely cannot ban (no adviceseeker/owner Matrix id). A matching `unbanUserFromRoom` primitive is included but `unbanUsersInChat` still runs the legacy mute path — wiring unban to the Matrix ban is a follow-up if we keep server-side bans.
2. **Member `status` / `utcOffset` are now `null`** in the chat-members response. Populating live presence per member would require a Synapse call per member (against the no-hot-loop rule); left out deliberately. Confirm the UI does not depend on these fields.
3. **`Messaging.isInChat(String, String)` removed from the port.** It had no external callers; the member check is now internal and Matrix-native.

## Stacking

Stacked on **#296**. Merge order: **#295 → #296 → this**. This branch already contains #296's commit, so the diff shown here is only the member-query sweep.

## Tests

TDD, per caller: happy path, RC-disabled behaviour, fail-safe on Matrix error, technical-account filtering. Full `./mvnw -B package` (JDK 21, no `testFailureIgnore`):

```
Tests run: 2227, Failures: 0, Errors: 0, Skipped: 7
BUILD SUCCESS
```

`spotless:apply` run before push.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/Messenger.java
+++ b/src/main/java/de/caritas/cob/userservice/api/Messenger.java
@@ -2,12 +2,14 @@ package de.caritas.cob.userservice.api;
 
 import static java.util.Objects.isNull;
 
+import de.caritas.cob.userservice.api.adapters.matrix.MatrixSynapseService;
 import de.caritas.cob.userservice.api.adapters.web.dto.AgencyDTO;
 import de.caritas.cob.userservice.api.model.Chat;
 import de.caritas.cob.userservice.api.model.Consultant;
 import de.caritas.cob.userservice.api.model.Session;
 import de.caritas.cob.userservice.api.model.Session.RegistrationType;
 import de.caritas.cob.userservice.api.model.Session.SessionStatus;
+import de.caritas.cob.userservice.api.model.User;
 import de.caritas.cob.userservice.api.port.in.Messaging;
 import de.caritas.cob.userservice.api.port.out.ChatRepository;
 import de.caritas.cob.userservice.api.port.out.ConsultantRepository;
@@ -16,6 +18,8 @@ import de.caritas.cob.userservice.api.port.out.SessionRepository;
 import de.caritas.cob.userservice.api.port.out.UserRepository;
 import de.caritas.cob.userservice.api.service.StringConverter;
 import de.caritas.cob.userservice.api.service.agency.AgencyService;
+import de.caritas.cob.userservice.api.service.matrix.GroupChatMembershipService;
+import de.caritas.cob.userservice.api.service.matrix.GroupChatMembershipService.ResolvedRoomMember;
 import java.time.LocalDateTime;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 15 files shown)


Shirloin:
1. Ban behavior: PR implements Matrix ban (stronger than RC mute — 
   blocks re-join). Is this the intended product behavior?

2. Unban: Matrix ban is not reversed on unban — unban still uses 
   RC path which is a no-op. Is this acceptable as a follow-up, 
   or must it be fixed before merge?

3. UI: Member status and utcOffset now return null for Matrix rooms. 
   Does the frontend handle null presence fields gracefully?

PR #306, opened 2026-07-05
test: delete dead Rocket.Chat adapter tests (RC teardown phase 2)

Storypapst:
## What & why

Rocket.Chat is deactivated in production (`rocket-chat.enabled=false`; the active bean is `DisabledRocketChatService`, registered via `@ConditionalOnProperty(havingValue="false", matchIfMissing=true)`). This PR removes three test files whose class-under-test is never reached on the Matrix-only path — they test a never-registered / never-called adapter.

Reference: **Test Quality Audit 2026-07-04 — RC teardown phase 2**.

## Deleted (with dead-proof)

| File | Class under test | Why dead |
|------|------------------|----------|
| `RocketChatServiceTest` (1364 lines) | `RocketChatService` | The real adapter is `@ConditionalOnProperty(name="rocket-chat.enabled", havingValue="true")` → **not a registered bean** in prod/default; `DisabledRocketChatService` replaces it. Pure Mockito test of code that never runs in production. |
| `RocketChatServiceIT` | `RocketChatService` (autowired) | `@SpringBootTest @ActiveProfiles("testing")` autowires `RocketChatService`, which in the testing profile resolves to the **disabled subclass** — so it asserts real-RC HTTP behaviour against no-op overrides. Also not in the surefire CI gate. |
| `RocketChatRollbackServiceTest` | `RocketChatRollbackService` | **Zero production callers** — `grep RocketChatRollbackService src/main` outside the class itself returns nothing. Dead code. |

**1860 lines of dead test code removed.**

## Deliberately KEPT (not deleted — still live)

- `RocketChatCredentialsProviderTest` — `RocketChatCredentialsProvider` is an **unconditional `@Service`**, still injected by `UnauthorizedMembersProvider`, `MessageServiceProvider`, `AskerImportService`, and `DisabledRocketChatService`. Still a registered, wired bean → its test stays. (This deviates from the audit's DELETE list; kept on the evidence that the bean is live.)
- `admin/service/rocketchat/*` operation-service tests (`RocketChatAddToGroupOperationServiceTest`, `RocketChatRemoveFromGroupOperationServiceTest`, `RocketChatOperationConditionProviderTest`) — those classes are still invoked by live assign/create facades (`AssignEnquiryFacade`, `AssignSessionFacade`, `RemoveConsultantFromRocketChatService`, `RocketChatAsyncHelper`, `RocketChatGroupOperation`). Their terminal RC calls no-op when RC is off, but the surrounding orchestration/condition logic still executes, so removing the coverage was judged too aggressive for this "safe deletions" PR.
- The RC-off contract guards — `RocketChatDisabledBootTest`, `DisabledRocketChatServiceTest`, `RocketChatConditionalRegistrationTest` — untouched (they assert the disabled behaviour; that is the RC-off contract).

## Verification

- `./mvnw test-compile` (JDK 21, `release 21`) → **BUILD SUCCESS** (376 test sources).
- No dangling references to the deleted classes anywhere in `src/test`.
- Remaining RC tests green after removal: `DisabledRocketChatServiceTest`, `RocketChatConditionalRegistrationTest`, `RocketChatCredentialsProviderTest`, `RocketChatFacadeTest`, `admin/service/rocketchat/*` → **53 tests, 0 failures**.

Test-only; no `src/main`, no schema, no `.github/` changes.


Storypapst:
Part of RC teardown phase 2 + C2. Companion PR #307 adds the C2 Matrix room-provisioning coverage (this PR is the safe deletions half). Coordinated with #300 (no file overlap).

PR #309, opened 2026-07-05
fix(adr008): data migration + inviteAndJoin hardening (follow-up #293)

Shirloin:
## Follow-up fixes for ADR-008 (#293)

Found during code review. Two fixes on top of the original ADR-008 commit:

**BLOCKER fix — data migration:**
Active session_supervisor rows created before ADR-008 still carry the 
client room id in matrix_room_id. Frontend reads this field to route 
supervisor asides — without this migration, asides for 12 existing 
supervisors still go to the client room, re-introducing the 
safeguarding leak #293 is fixing.

**WARNING fixes — inviteAndJoin hardening:**
- Throw InternalServerErrorException when loginAsUserAccessToken 
  returns null instead of silently skipping joinRoom
- Handle "already in the room" on invite to prevent brittle 500s 
  on re-add after partial removeSupervisor failure


shazia-k (approved):

shazia-k (commented):
Check the Java build field formatting error.

Shirloin:
Spotless/formating is fixed on 2b11e45, main CI is green.
The red job is the non-blocking integration-test burn in, unrelated to this PR.

PR #324, opened 2026-07-07
fix(case-handover): seed reason-policy table on dev; harden policy update

kiodreambau:
## Symptom (reported by @shazia-k, dev, 2026-07-07)

Under **Settings → Permissions → Case handover**, the toggles **Enabled** (module master) and the per-reason **client-consent** switches have no effect: the switch flips back right after clicking, no change persists.

## Diagnosis

Verified against the PreDev cluster: every `PUT /service/users/case-handover/reason-policies` returns **HTTP 500** (empty body, ~10 ms), while `GET` returns 200. The admin card applies an optimistic update and rolls back on error — hence "no effect after click". Both reported toggles go through this one endpoint.

The 500 is **environment state, not the request**:

- The request body is well-formed (screenshot from @shazia-k shows a complete 5-element payload) — a bad body would be **400**, not 500.
- A DB constraint violation maps to **409** in this service — not what we see.
- `spring.jpa.hibernate.ddl-auto=validate` passes at startup, so the dev table schema **matches** the entity (no missing/renamed column).
- Liquibase runs as the **application datasource user** (no separate liquibase user), so the app user **can write** — grants are not the issue.

Root cause: changeset **`0057`** guarded *both* `case_handover_request` *and* `case_handover_reason_policy` behind a single `AND` precondition with `onFail="MARK_RAN"`. On dev, `case_handover_request` had been **hand-applied** before `0057` ever ran (see the changeset's own comment), so the precondition failed and the reason-policy `CREATE` **and** seed were skipped wholesale. `GET` still returns 200 because `listReasonPolicies()` silently falls back to a hardcoded default list when the table is empty — masking the unseeded state until the first write.

## Changes

- **`0059` repair changeset** (`context="seed"`, idempotent): ensures `case_handover_reason_policy` exists and seeds the default reasons, **decoupled** from `case_handover_request`. `CREATE TABLE IF NOT EXISTS` + `INSERT ... ON DUPLICATE KEY UPDATE` → no-op where the table is already present/seeded, never clobbers admin edits.
- **`updateReasonPolicies` hardening**: collect existing policies with a merge function so duplicate stored codes can't raise `IllegalStateException` (→ 500).
- **Regression test** for the duplicate-code path.

## Verification (needs a dev redeploy)

The fix runs on the next userservice deploy to dev (liquibase applies `0059`). After deploy, please re-test in the admin: toggling **Enabled** / a client-consent switch should now `200` and persist across reload.

> Note on confidence: the write-path 500 is logged server-side as an internal error, but the dev pod only retains ~10 min of logs, so the exact stack trace of the earlier failures had already rolled off and could not be captured. The diagnosis above rules out the parse/constraint/schema/grant causes; the remaining cause consistent with all evidence is the unseeded table from the skipped `0057`. If post-deploy testing still shows a 500, grab the fresh stack trace (`kubectl -n caritas logs -f <userservice-pod>` while clicking once) and we adjust.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/service/CaseHandoverService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/CaseHandoverService.java
@@ -144,7 +144,9 @@ public class CaseHandoverService {
         caseHandoverReasonPolicyRepository.findAllByOrderByDisplayOrderAscCodeAsc().stream()
             .collect(
                 Collectors.toMap(
-                    CaseHandoverReasonPolicy::getCode, Function.identity(), (first, ignored) -> first));
+                    CaseHandoverReasonPolicy::getCode,
+                    Function.identity(),
+                    (first, ignored) -> first));
     LocalDateTime now = LocalDateTime.now();
     List<CaseHandoverReasonPolicy> policiesToSave =
         requestedReasons.stream()
--- a/src/main/java/de/caritas/cob/userservice/api/service/CaseHandoverService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/CaseHandoverService.java
@@ -137,9 +137,14 @@ public class CaseHandoverService {
       throw new BadRequestException("At least one handover reason policy is required");
     }
 
+    // Keep the first row per code: a plain toMap throws IllegalStateException (-> 500)
+    // if the table ever holds duplicate codes, e.g. after a hand-applied seed on an
+    // environment where the guarded 0057 changeset was skipped.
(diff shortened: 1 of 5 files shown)


shazia-k (commented):
Key risk notes
The 0059 changeset is tagged context="seed" — it never runs on production.
The service hardening is backward-compatible and a no-op when all codes are unique.
One follow-up recommendation: add a UNIQUE KEY constraint on code in a future changeset to enforce uniqueness at the DB level and prevent silent masking.

PR #106, opened 2026-06-13
refactor: split RocketChatService into 7 focused client classes (#93)

einsteinjava:
## Summary

Split `RocketChatService.java` (1,402 lines, 67 methods) into 8 focused files using the **Delegation pattern**. The original class is now a ~530-line facade that delegates to 6 new client classes.

**Issue:** #93

## Before → After

| Before | After |
|---|---|
| 1 file, 1,402 lines, 67 methods | 9 files, 25-598 lines each |
| 13 distinct responsibilities | 1 responsibility per file |
| 7 injected dependencies | 3-6 per client (only what it needs) |

## New Files (8)

| File | Lines | Purpose |
|---|---|---|
| RocketChatEndpoints.java | 43 | 26 shared API endpoint constants |
| RocketChatHttpHeaders.java | 25 | Auth header builder |
| client/RocketChatUserClient.java | 412 | User CRUD, login/logout, user search |
| client/RocketChatGroupClient.java | 598 | Group CRUD, members, paginated listing |
| client/RocketChatRoomClient.java | 198 | Room info, chat users (MongoDB), room settings |
| client/RocketChatMessageClient.java | 96 | Message cleanup |
| client/RocketChatPresenceClient.java | 94 | User presence/availability |
| client/RocketChatSubscriptionClient.java | 105 | Subscriptions, E2E key update |

## Safety

- ✅ No behavior/API/DB changes
- ✅ 59/59 RocketChatServiceTest pass
- ✅ RocketChatRollbackService unchanged (verified)
- ✅ `@Cacheable` stays on facade
- ✅ Full test suite: same baseline as before refactor

PR #67, opened 2026-06-08
MT-04-08 support multiple tenant admins

radaljj:
Implements MT-04-08 backend changes for multiple Tenant-Admins per Tenant-Unit.

Changes:

Add UPDATE_DATE sort support for Tenant-Admin search.
Stop assigning legacy single-tenant-admin role to newly created Tenant-Admins.
Allow Tenant-Admin search/list to continue when old dev data references a missing tenant.
Add targeted backend tests.
Verified locally:

Created multiple Tenant-Admins with the same tenant_id.
Deleted one Tenant-Admin and confirmed the other remained.
Verified UPDATE_DATE sorting via backend/API for ASC and DESC.
Open:

Last Tenant-Admin removal rule still needs product clarification.

PR #328, opened 2026-07-07
ci: non-blocking oasdiff API contract diff on PRs

kiodreambau:
## What

Adds a non-blocking `api-contract-diff` job to `ci-pull-request.yml`: it checks out the PR base branch into a subdirectory and runs the `tufin/oasdiff` docker image (`breaking`, `--fail-on WARN`) over every vendored own-API spec (`api/userservice.yaml`, `api/useradminservice.yaml`, `api/conversationservice.yaml`, `api/appointmentservice.yaml`, `api/userstatisticsservice.yaml`) against the PR head. `continue-on-error: true` keeps it informational until the team agrees on a process for intentional breaking changes — same pattern as AgencyService #106, TenantService #67, ConsultingTypeService #51.

## Verification

Local run with the docker image: identical specs → exit 0 (`No changes detected`); diffing two different specs → exit 1 with `[api-path-removed-without-deprecation]` errors, so both the pass and fail paths work.

### DIFF
--- a/.github/workflows/ci-pull-request.yml
+++ b/.github/workflows/ci-pull-request.yml
@@ -46,3 +46,39 @@ jobs:
       uses: ./.github/actions/maven-verify-burnin
       with:
         java_version: '21'
+
+  # Non-blocking OpenAPI contract diff: flags breaking changes to this service's
+  # own vendored specs (api/*.yaml) between the PR base and head. Informational
+  # for now; flip to blocking once the process around intentional breaking
+  # changes (version bump / consumer sign-off) is agreed.
+  api-contract-diff:
+    name: api-contract-diff (non-blocking)
+    runs-on: ubuntu-latest
+    continue-on-error: true
+
+    steps:
+    - name: Checkout PR head
+      uses: actions/checkout@v6
+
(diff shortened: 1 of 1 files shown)


kiodreambau:
Heads-up on the first run of the new job (red as expected, non-blocking): 4 of 5 specs pass with `No changes detected`, but `api/appointmentservice.yaml` fails to **load** in oasdiff:

```
Error: failed to load base spec from "/work/.base/api/appointmentservice.yaml": map key "LanguageCode" not found
```

That's a pre-existing defect in the vendored spec itself: line 267 references `#/components/schemas/LanguageCode`, but no such schema is defined in the file (reproducible locally with `docker run --rm -v "$PWD:/work:ro" tufin/oasdiff breaking /work/api/appointmentservice.yaml /work/api/appointmentservice.yaml`). The OpenAPI generator is apparently more lenient than oasdiff's parser.

So the red check is signal, not a job bug — the job will go green once the missing `LanguageCode` schema is added to (or the dangling `$ref` removed from) `api/appointmentservice.yaml`. Happy to do that here if maintainers prefer a green first run, but I kept this PR CI-only.

PR #342, opened 2026-07-08
Propagate consultant agency assignment validation errors

### LINKED ISSUE #341: Consultant agency assignment returns 200 after validation failure
## Problem

`UserAdminController.setConsultantAgencies` catches assignment exceptions and still returns HTTP 200. When `ConsultantTopicAgencyCompatibilityValidator` rejects an assignment, the Admin UI believes the consultant agency assignment succeeded even though the backend did not apply it correctly.

Observed on 2026-07-08 during Caritas demo / PreDev consultant assignment debugging. Example validation message:

```
Consultant topic ids [10] are not covered by selected/assigned agencies [272]
```

## Expected

Validation failures propagate as HTTP 400 and expose the validation message in the response body so the Admin UI can show a real error.

Assignments that already exist remain idempotent: existing consultant-agency relations are skipped before creating new relations.

## Actual

The controller catches broad `Exception`, logs with `System.out.println`, and always returns HTTP 200.

## Scope

Fix the controller behavior and add regression tests. Do not change the existing delete/replace behavior in this PR because that path was previously marked as causing 403 during Matrix migration.

Storypapst:
## Problem

`UserAdminController.setConsultantAgencies` swallowed every assignment exception and still returned HTTP 200. That made the Admin UI report successful consultant-agency assignment even when `ConsultantTopicAgencyCompatibilityValidator` rejected the request.

Fixes #341.

## Solution

- Remove the broad catch/`System.out.println` path from `setConsultantAgencies`.
- Check permissions, copy the request list, filter already persisted agency relations, then run the existing prepare/complete assignment flow.
- Let `BadRequestException` propagate as HTTP 400.
- Return the `BadRequestException` message as `{"message":"..."}` so the Admin UI can display the backend validation reason.

## Notes

This keeps already assigned agencies idempotent by filtering them before creation. It intentionally does not restore the full OpenAPI replace/delete semantics because the previous delete path is documented in the controller as causing a 403 during Matrix migration.

The existing validator tests already document the bootstrap decision that offline agencies count for topic coverage, so newly created offline agencies can receive their first consultant.

## Validation

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -q spotless:check`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -q -Dspotless.check.skip=true -Dtest=UserAdminControllerIT,ApiResponseEntityExceptionHandlerTest,ConsultantAdminFacadeTest,ConsultantTopicAgencyCompatibilityValidatorTest test`
- `git diff --check`

## Review

Do not merge automatically. Frank reviews this PR before merge/deploy.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserAdminController.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserAdminController.java
@@ -44,6 +44,7 @@ import io.swagger.annotations.Api;
 import jakarta.validation.Valid;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.List;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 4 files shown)


Storypapst:
Closing as superseded by #343 as the main implementation. The missing BadRequest response-body contract from this PR was moved into follow-up #345 against #343.

PR #345, opened 2026-07-08
Expose BadRequest messages in PR 343

Storypapst:
## Summary

Follow-up for #343.

PR #343 is the better main fix for consultant agency assignment because it keeps the set semantics in `ConsultantAdminFacade`. This small follow-up keeps the missing response contract from #342: semantic `BadRequestException` failures now return a stable `{ "message": "..." }` body so Admin can surface the backend validation text.

## Validation

- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -q -Dspotless.check.skip=true -Dtest=ApiResponseEntityExceptionHandlerTest,UserAdminControllerIT,ConsultantAdminFacadeTest test`
- `JAVA_HOME=$(/usr/libexec/java_home -v 21) ./mvnw -q spotless:check`
- `git diff --check`


shazia-k (approved):
This looks good

Issue #91, opened 2026-06-11
Backend refactoring: reduce service coupling and large class responsibilities


radaljj:
### Target Service/Component

User Service

### Priority

P2-Medium

### Problem Statement

@Storypapst 
Several backend services are structured as microservices, but the current codebase still contains large classes with mixed responsibilities.

Initial analysis shows that ORISO-UserService is the main coupling point and contains the largest maintainability risks.

The main hotspots are:

UserController
RocketChatService
MatrixSynapseService
SessionService
KeycloakService
These classes combine multiple business and integration flows, making future changes harder to review, test and maintain.

### Proposed Solution

Create a dedicated backend refactoring track focused on reducing class size, improving responsibility boundaries and making the code easier to maintain.

This track should:

keep all behavior unchanged
avoid API contract changes
avoid database changes
split work into small reviewable PRs
start with ORISO-UserService because it has the highest coupling
create smaller follow-up issues for each major hotspot

### Alternatives Considered

A large one-time refactor was considered, but it would be too risky and difficult to review.

Another option is to leave the structure unchanged and only fix bugs, but that keeps the same maintainability problems and slows down future work.

### Impact & Benefits

Target users:

backend developers
reviewers
QA indirectly, through lower regression risk

Expected benefits:
easier code navigation
smaller and safer PRs
better testability
lower risk when changing user/session/chat flows
clearer ownership of controller, service, facade and adapter responsibilities

### Technical Considerations

Affected services:

ORISO-UserService primarily
other backend services may be reviewed later if similar issues are found
Database changes needed:

No
API changes needed:

No
Security considerations:

Authentication and authorization behavior must remain unchanged
Performance considerations:

No performance impact expected
Refactor should preserve existing runtime behavior

### Related Documentation/Links

Related follow-up issues:

Refactor ORISO-UserService UserController responsibilities
Refactor ORISO-UserService RocketChatService responsibilities
Refactor ORISO-UserService MatrixSynapseService responsibilities
Refactor ORISO-UserService SessionService responsibilities
Refactor ORISO-UserService KeycloakService responsibilities

### Checklist

- [x] I have searched existing issues to ensure this feature hasn't been requested before
- [x] I have provided a clear problem statement and proposed solution
- [x] This feature aligns with ORISO's mission of accessible mental health support

PR #346, opened 2026-07-08
Extract Keycloak auth/session ops into KeycloakAuthClient.

radaljj:
## Summary
First slice of the Keycloak refactor (#91).

Pulls login, logout, OTP-ignorant password verify, and session close out of `KeycloakService` into a dedicated `KeycloakAuthClient`, so the auth/session surface stays readable and reviewable on its own.

`KeycloakService` remains the `IdentityClient` facade and simply delegates — no API or behavior change.

## Test plan
- [x] `KeycloakServiceTest` + `KeycloakServiceLoggingTest`
- [ ] CI green

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakAuthClient.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakAuthClient.java
@@ -1,0 +1,175 @@
+package de.caritas.cob.userservice.api.adapters.keycloak;
+
+import static java.util.Objects.nonNull;
+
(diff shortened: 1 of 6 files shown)


shazia-k (commented):
2 medium issues to address before merge:
SensitiveKeycloakFormData — change from package-private static final class back to private static final class (visibility was widened unintentionally during the move).
No dedicated KeycloakAuthClientTest — the new class owns all OIDC auth logic and has edge-case gaps (verifyIgnoringOtp with null body / null refreshToken, wasLogoutSuccessful with non-204 non-exception response).

shazia-k (approved):

Issue #258, opened 2026-07-01
unit tests: ChatService — increase coverage


Kadyan25:
**Type:** Test coverage

## Problem
ChatService currently sits at 72% instruction coverage with existing tests. Several branches and paths are not exercised.

## Fix
Extend ChatServiceTest with targeted tests covering uncovered branches to reach 80%+ instruction coverage.

## Verification
- [ ] Run JaCoCo on ChatServiceTest — instruction coverage ≥ 80%

Storypapst:
Resolved via PR#260 (merged 2026-07-02). Closing as part of the GitHub hygiene sweep — issue was already fixed but never marked closed.

PR #347, opened 2026-07-08
fix(userservice): don't fail consultant login when no agency is assigned

kiodreambau:
## Problem

Focused on the reported symptom: **a consultant cannot log into the app.** Authentication actually succeeds — the token is issued with the `consultant` role — but the very first call the app makes afterwards, `GET /users/data`, returned **500** and dropped the consultant on `error.404.html`, so the account looked unusable.

Root cause (verified on PreDev, correlationId in server logs):

```
InternalServerErrorException: No agency available for consultant ed6ff102-...
  at ConsultantDataProvider.retrieveData(ConsultantDataProvider.java:57)
```

`retrieveData` threw whenever the consultant had **no assigned agency**. On PreDev this hit every consultant whose agency assignment had silently failed (the swallowed-error bug fixed in #343), turning "no agency yet" into "can't log in at all".

## Change

Return the consultant profile with an **empty agency list** instead of throwing. This is consistent with the rest of the same provider, which already degrades gracefully — `agencyDTOsOf()` catches agency-service errors and returns an empty list rather than blocking login, with the comment *"Do not block login when agency-service cannot provide metadata"*. The lone hard `throw` on empty agencies contradicted that intent.

Downstream fields already handle empty agencies safely (`hasAnonymousConversations` → `false`, `hasArchive` reads the session repo).

## Tests

- `ConsultantDataProviderTest`: the former `..._Should_ThrowInternalServerErrorException_When_NoAgenciesFound` becomes `..._Should_ReturnMinimalProfile_When_NoAgenciesFound`, asserting an empty agency list and a populated profile.
- `./mvnw test -Dtest=ConsultantDataProviderTest` → 11/11 green.

## Context

The durable cure for consultants ending up agency-less is #343 (agency assignment no longer swallows validation errors) + ORISO-Admin #258 (surfaces them), both already merged to `dev`. This PR is the safety net so a consultant temporarily without an agency can still load the app instead of being locked out.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/facade/userdata/ConsultantDataProvider.java
+++ b/src/main/java/de/caritas/cob/userservice/api/facade/userdata/ConsultantDataProvider.java
@@ -9,7 +9,6 @@ import de.caritas.cob.userservice.api.adapters.web.dto.AgencyDTO;
 import de.caritas.cob.userservice.api.adapters.web.dto.EmailToggle;
 import de.caritas.cob.userservice.api.adapters.web.dto.EmailType;
 import de.caritas.cob.userservice.api.adapters.web.dto.UserDataResponseDTO;
-import de.caritas.cob.userservice.api.exception.httpresponses.InternalServerErrorException;
 import de.caritas.cob.userservice.api.helper.AuthenticatedUser;
 import de.caritas.cob.userservice.api.manager.consultingtype.ConsultingTypeManager;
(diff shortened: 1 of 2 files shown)


radaljj (approved):

kiodreambau:
CI green, already approved by @radaljj. @Hassan9215 requesting your review before merge.

radaljj (approved):

PR #370, opened 2026-07-09
test: lock Matrix server_name invariants (ADR-005 / DB-M04)

kiodreambau:
## Why

ADR-005 rebuilds the Matrix homeserver with `server_name = matrix.oriso.org`. The `server_name` is immutable inside every Matrix user ID (`@alice:matrix.oriso.org`) and room ID (`!abc:matrix.oriso.org`), so it used to bake a bare Hetzner IP (`91.99.219.182` / `91.99.183.160`) into every MXID — finding **DB-M04**. This PR adds regression tests that lock the invariants which keep that class of bug from silently coming back. No production behaviour changes.

Part of the ADR-005 homeserver rebuild tracked in the epic `OpenResilienceInitiative/ORISO-Helm#31`.

## What (tests only)

- **`MatrixIdsTest`** (extended) — edge cases documenting current `MatrixIds` parsing: server parts with dots and ports, bare IPv4 servers (with/without port), empty local part, colon-first input that throws, lenient-variant behaviour, and `isUserId` / `isRoomId` sigil detection being independent of whether the server part is a DNS name or an IP.
- **`AssignEnquiryFacadeTest`** (extended) — proves the one place the service constructs an MXID (`AssignEnquiryFacade`, the `"@" + localpart + ":" + matrixConfig.getServerName()` fallback) derives its server part from the injected `MatrixConfig.serverName`. A distinctive `test.example.org` is injected (not the production value) and asserted to appear verbatim in the constructed MXID — change the config, change the MXID.
- **`MatrixConfigPropertiesTest`** (new) — asserts `matrix.serverName` / `matrix.apiUrl` in `application.properties` resolve from `MATRIX_SERVER_NAME` / `MATRIX_API_URL` with **no** hardcoded, non-empty default.
- **`NoHardcodedMatrixServerIpTest`** (new) — source guard scanning `src/main`; fails if any file hardcodes an IPv4 as a Matrix server name / homeserver URL. Deliberately precise (only IPs in a Matrix/Synapse/homeserver context, plus the known former homeserver IPs), and tolerant of loopback/wildcard addresses and the explanatory `# NOT hardcoded IPs like ...` documentation comments.

## Red/green discipline

Every assertion was proven capable of failing by temporarily mutating production code/config, then reverted:

| Mutation | Test that caught it |
| --- | --- |
| `MatrixIds.localpart` `indexOf(':')` → `lastIndexOf(':')` | `MatrixIdsTest` (dotted-server / IPv4-with-port cases) |
[...]

PR #372, opened 2026-07-10
fix(build): stop surefire running the unit tests twice

kiodreambau:
## Every `mvn test` runs the whole unit suite twice

`maven-surefire-plugin` declares an explicit `unit-tests` execution bound to the `test` phase, but never disables surefire's **implicit `default-test` execution**, which is bound to the same phase. Both run.

A second, quieter bug falls out of the same cause: `${skip.unit-tests}` is configured only on the `unit-tests` execution, so the implicit one ignores it — **`-Dskip.unit-tests=true` did not skip anything.**

## Fix

Disable `default-test` with `<phase>none</phase>` and let `unit-tests` own the phase. The `integration-tests` execution (bound to `integration-test`) is untouched.

## Nothing is excluded

`default-test`'s default includes are `**/Test*.java`, `**/*Test.java`, `**/*Tests.java`, `**/*TestCase.java`; `unit-tests` includes only `**/*Test.java`. The difference is therefore `Test*.java`, `*Tests.java`, `*TestCase.java`. In this repo:

| pattern | count |
|---|---|
| `*Test.java` | 329 |
| `*Tests.java` | 0 |
| `*TestCase.java` | 0 |
| `Test*.java` (not `*Test.java`) | 4 |

Those four are `TestMailsControllerApi`, `TestAgencyControllerApi`, `TestConstants`, `TestLogAppender` — **helper classes with zero `@Test` methods**. So no executable test is lost.

## Verification (Java 21, on `pre-dev`, scoped to `MatrixIdsTest,AssignEnquiryFacadeTest`)

| | before | after |
|---|---|---|
| per-class `Tests run:` summary | printed **2×** each | printed **1×** each |
| aggregate | `Tests run: 45` **2×** | `Tests run: 45` **1×** |
| per-class counts | 19 + 26 | **19 + 26 (unchanged)** |
| `-Dskip.unit-tests=true` | still ran the class (**1×**) | correctly skips (**0×**) |

Measured by restoring the unpatched `pom.xml` for the "before" run so both runs are on the same tree.

## Notes
- Pre-existing; found incidentally while adding the ADR-005 tests (PR #370). Unrelated to that change.
- The repo targets **Java 21** and no JDK is on `PATH` by default (`JAVA_HOME=$(/usr/libexec/java_home -v 21)`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1175,6 +1175,14 @@
 				<groupId>org.apache.maven.plugins</groupId>
 				<artifactId>maven-surefire-plugin</artifactId> <!-- surefire plugin version managed by Spring Boot -->
 				<executions>
+					<!-- Surefire binds an implicit `default-test` execution to the `test` phase.
+					     Left enabled it runs the unit tests a SECOND time alongside the explicit
+					     `unit-tests` execution below, and it ignores `${skip.unit-tests}`.
+					     Disable it so `unit-tests` solely owns the phase. -->
+					<execution>
+						<id>default-test</id>
+						<phase>none</phase>
+					</execution>
 					<execution>
 						<id>unit-tests</id>
 						<phase>test</phase>


radaljj (approved):

radaljj (changes requested):
Please retarged to pre-dev  branch instead of dev

shazia-k (approved):

PR #384, opened 2026-07-11
fix: allow consultants to accept registered enquiries

Storypapst:
Closes #383

## Problem
A consultant could list registered enquiries but received HTTP 403 when accepting a NEW enquiry. The CONSULTANT role had `ASSIGN_CONSULTANT_TO_SESSION`, while the NEW-enquiry delegate correctly checks `ASSIGN_CONSULTANT_TO_ENQUIRY`.

## Change
- grant `ASSIGN_CONSULTANT_TO_ENQUIRY` to the CONSULTANT role
- extend the authority mapping regression test

## TDD evidence
- RED: `AuthorityTest` failed on the missing enquiry authority
- GREEN: focused authority/delegate tests: 37/37 passed
- required test suite: 3,061 tests passed, 0 failures/errors, 7 skipped
- package build: passed

## Known non-blocking burn-in
`mvn verify` reproduces the repository-wide integration-test baseline (911 ITs; 240 failures, 266 errors). This job is explicitly `continue-on-error` in the verify burn-in action and the failures span unrelated 401/application-context/lazy-loading/locking scenarios; the required `mvn test` gate and package build are green.

## PreDev QA after approval
Deploy the regular UserService image, then accept registered session 103139 as a real consultant and prove encrypted two-way messaging plus reload.

Issue #383, opened 2026-07-11
[ADR-006-US] Allow consultants to accept registered enquiries


Storypapst:
## Reproduction

On PreDev, consultant A can list the fresh registered Agency Counselling enquiry `103139` (`GET .../enquiries/registered` → `200`) but assigning it to the same consultant returns `403`.

## Root cause

`Authority.CONSULTANT` grants `ASSIGN_CONSULTANT_TO_SESSION`. The assign endpoint security accepts either session or enquiry authority, but `UserSessionControllerDelegate` requires `ASSIGN_CONSULTANT_TO_ENQUIRY` specifically when `SessionStatus.NEW`. No normal consultant mapping currently grants that authority.

## Acceptance

- [ ] Add a failing authorization/authority regression test for a consultant accepting a NEW registered enquiry.
- [ ] Grant the minimum required enquiry authority to the correct consultant role without broadening admin permissions.
- [ ] Full relevant UserService gate passes.
- [ ] Deploy the regular PreDev image.
- [ ] Consultant accepts registered enquiry `103139` and encrypted two-way Agency Counselling messages plus reload pass.

Part of OpenResilienceInitiative/ORISO-Frontend#408 and #410.

Storypapst:
PreDev QA passed on regular image digest `sha256:25fa2898c874453e4a5a23a57f35f1b915fc4afa3de183efaa15a1f838cbecef`.

Evidence:
- clean NEW registered session 103149 accepted with HTTP 200
- UserService logs agency invite and successful consultant join to Matrix room `!PEeyxgjvgrOQzbDtFb:91.99.183.160`
- encrypted asker → consultant Matrix send: 200
- encrypted consultant → asker Matrix send: 200
- both clear messages remain visible after asker reload

Fix chain: #384 (authority), #385 (disabled Rocket.Chat empty-member path), #386 (route NEW enquiries through the existing Matrix assignment facade).

Storypapst:
## Bug Fix Run screenshot evidence

A consultant successfully accepts the registered enquiry on PreDev.

![Accepted registered enquiry](https://github.com/OpenResilienceInitiative/ORISO-E2E/blob/719a815dd6178d8593604c52542a224b302b5767/docs/e2e-runs/e2e-20260719-1711/screenshots/20-consultant-accepted-enquiry-predev-desktop.png?raw=1)

PR #385, opened 2026-07-11
fix: skip Rocket.Chat cleanup without members

Storypapst:
Continues #383

## Deployed reproduction
After #384 removed the incorrect 403, accepting registered session 103139 reached assignment logic but returned HTTP 500. The deployed stack trace points to `UnauthorizedMembersProvider.addTechnicalUsers`: with Rocket.Chat disabled, the adapter returns an empty member list, yet the provider still requires the absent Rocket.Chat technical user.

## Minimal fix
When the group contains no Rocket.Chat members, return an empty removal list before resolving any Rocket.Chat technical identity. There is nobody to remove, and the Matrix-only path no longer depends on removed Rocket.Chat credentials.

## TDD evidence
- RED: empty-member regression test failed while resolving the technical user
- GREEN: provider + assignment-facade tests pass
- full required suite: 3,062 tests, 0 failures/errors, 7 skipped
- package build: passed

## PreDev QA
Deploy the regular image and retry session 103139 acceptance, then prove encrypted asker/consultant two-way messaging and reload.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/facade/assignsession/UnauthorizedMembersProvider.java
+++ b/src/main/java/de/caritas/cob/userservice/api/facade/assignsession/UnauthorizedMembersProvider.java
@@ -47,6 +47,9 @@ public class UnauthorizedMembersProvider {
       Consultant consultant,
       List<GroupMemberDTO> memberList,
       Consultant consultantToKeep) {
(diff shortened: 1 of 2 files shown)


Storypapst:
@coderabbitai review

PR #390, opened 2026-07-11
feat: tenant-scoped admin dashboard statistics endpoint

Storypapst:
## Problem

The admin statistics dashboard (ORISO-Admin) shows mock data only — no backend endpoint exists that serves real counselling statistics. KDG compliance requires statistics to come from the application-layer database (never from monitoring) with re-identification protection.

## What changed

- `GET /useradmin/statistics/dashboard` (+ `/service/...` variant): caller-scoped statistics — platform admin → per-tenant rows, tenant admin → tenant total + per-agency rows, agency admin → own agencies only (via `AdminAgency`)
- `AdminStatisticsRepository`: native aggregate queries with explicit tenant restriction (Hibernate tenant filters do not apply to native SQL)
- Small-cell suppression: counselling metrics are omitted (`suppressed: true`) when a target covers fewer than 2 counsellor accounts
- Metrics: monthly enquiries (current/previous), active cases, active conversations today, group chats, daily new-session series (5 weeks), monthly top topics with share (4 months), session/chat counts per period
- SecurityConfig: endpoint restricted to `USER_ADMIN`, `TENANT_ADMIN`, `SINGLE_TENANT_ADMIN`, `RESTRICTED_AGENCY_ADMIN`

## Verification

- `mvnw test -Dtest=AdminDashboardStatisticsServiceTest` — 8/8 green (scope resolution, suppression, topic shares, daily series)
- `AdminStatisticsRepositoryIT` (@DataJpaTest) — every native query executes against the H2 testing schema
- `UserServiceApplicationIT` — context boot green
- `spotless:check` green

Frontend counterpart: ORISO-Admin PR (linked below).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #391, opened 2026-07-11
feat: consultant-scoped profile statistics endpoint

Storypapst:
## Why

The consultant profile statistics tab in ORISO-Frontend calls `/service/statistics/consultant`, an endpoint of the upstream Caritas **statisticsservice — which does not exist in ORISO**. The call 404s whenever the tab is enabled (`featureStatisticsEnabled` or non-tenant context).

## What

A consultant-scoped statistics endpoint in UserService, following the pattern of the admin dashboard statistics (#390):

- `GET /users/statistics/consultant` (+ `/service` prefix), query params `startDate`/`endDate` (`yyyy-MM-dd`, end inclusive)
- **KDG scope:** a consultant only ever sees their **own** numbers — the consultant id is taken from the authenticated user, never from the request; `CONSULTANT_DEFAULT` authority required
- **App-layer DB only:** `numberOfAssignedSessions` (sessions assigned to the consultant, created in the period) and `numberOfActiveSessions` (assigned, in progress, touched in the period)
- Upstream metrics without an app-layer source (**sent messages, video call duration, appointments**) are intentionally dropped from the contract

## Notes

- Assignment timestamps are not recorded in the app layer, so `numberOfAssignedSessions` uses the session creation date as the period anchor (documented in the repository).
- Frontend counterpart PR (tab rework + endpoint repoint) follows in ORISO-Frontend.

## Tests

- `ConsultantStatisticsServiceTest` — role guard, date validation, exclusive upper bound, own-id scoping (5 tests)
- `ConsultantStatisticsRepositoryIT` — native queries execute against the H2 testing schema

All green locally on JDK 25; spotless applied.


Storypapst:
**Verified live on Pre-Dev** (hot-deployed this branch as `oriso-userservice:consultant-stats`, image import via k3s ctr, additive to current `:dev`):

| Case | Result |
|---|---|
| Unauthenticated, via ingress `api.oriso-dev.site/service/users/statistics/consultant` | **401** |
| Unauthenticated, pod-direct `/users/statistics/consultant` | **401** |
| Consultant token (`e2e_consultant_a_20260710_1922`, tenant 82), 2026-07-01..11 | **200** `{"numberOfAssignedSessions":12,"numberOfActiveSessions":2}` |
| Same window counted directly in MariaDB | **12 / 2 — exact match** |
| Asker token (`bright_fish_ellis_643`) | **403** (KDG own-numbers guard) |
| `startDate > endDate` | **400** |
| Malformed date (`01.07.2026`) | **400** |

Ingress rewrite `/service/users(/|$)(.*) → /users/$2` lands exactly on the controller mapping, so both public and in-cluster paths work.

Rollback if needed: `kubectl set image deploy/oriso-platform-userservice -n caritas userservice=ghcr.io/openresilienceinitiative/oriso-userservice@sha256:9fcc842c6d023eb79c783f04808b72c40e2b3f4f6ffc13d375134cdd86e1672a` (+ imagePullPolicy back to default). The hot deploy is temporary anyway — the next `deploy-userservice-dev.sh` overwrites it.

radaljj (changes requested):
Before merge:

1. **Tenant on native queries** — match #390 / `AdminStatisticsRepository`: Hibernate `tenantFilter` does not apply to native aggregations. Please add an explicit `tenant_id` predicate (from `TenantContext`) when multitenancy is on.

2. **Role guard vs authority** — `SecurityConfig` allows any `CONSULTANT_DEFAULT` (including `group-chat-consultant` and `supervisor-consultant`), but `isConsultant()` only accepts the `consultant` role, so those users get 403 after auth. Drop the redundant service guard or align it with the authority/roles that already pass security.

3. **Retarget base to `pre-dev`** (current team policy).

Nits: OpenAPI yaml not updated; no Authorization IT; assignment metric uses `create_date` as the period anchor (documented — fine).

Storypapst:
Addressed the two blocking review points:

1. Tenant scoping on native queries -- both ConsultantStatisticsRepository queries now carry an explicit tenant_id predicate, resolved the same way as AdminDashboardStatisticsService (#390): TenantContext.getCurrentTenant() falling back to authenticatedUser.getTenantId().
2. Role guard vs authority mismatch -- the service-level guard now checks for the CONSULTANT_DEFAULT granted authority (same gate SecurityConfig already enforces) instead of the narrower consultant role literal, so group-chat-consultant/supervisor-consultant callers no longer 403.
3. Base branch retargeted to pre-dev per current team policy.

Extended unit tests to cover both fixes. All local tests green on JDK 25, spotless applied on JDK 21.

Left the OpenAPI-yaml and Authorization-IT nits as-is, consistent with #390 which didn't add either.

@radaljj could you take another look when you have a moment?

radaljj:
> Addressed the two blocking review points:
> 
> 1. Tenant scoping on native queries -- both ConsultantStatisticsRepository queries now carry an explicit tenant_id predicate, resolved the same way as AdminDashboardStatisticsService ([feat: tenant-scoped admin dashboard statistics endpoint #390](https://github.com/OpenResilienceInitiative/ORISO-UserService/pull/390)): TenantContext.getCurrentTenant() falling back to authenticatedUser.getTenantId().
> 2. Role guard vs authority mismatch -- the service-level guard now checks for the CONSULTANT_DEFAULT granted authority (same gate SecurityConfig already enforces) instead of the narrower consultant role literal, so group-chat-consultant/supervisor-consultant callers no longer 403.
> 3. Base branch retargeted to pre-dev per current team policy.
> 
> Extended unit tests to cover both fixes. All local tests green on JDK 25, spotless applied on JDK 21.
> 
> Left the OpenAPI-yaml and Authorization-IT nits as-is, consistent with #390 which didn't add either.
> 
> @radaljj could you take another look when you have a moment?

@Storypapst Re-checked the follow-up changes.

My original blocking points are resolved:
- native aggregate queries now include an explicit tenant_id predicate
- tenant resolution follows the TenantContext -> authenticatedUser fallback
- the service guard now checks CONSULTANT_DEFAULT, matching SecurityConfig
- the PR is now targeted to pre-dev

Frank’s Pre-Dev smoke results also cover the expected 401/403/200 and invalid-date cases.

One deployment note: the new message-count work requires STATISTICS_MESSAGE_COUNT_HMAC_SECRET to be configured, because ConsultantIdentityHasher fails fast when it is blank. Please make sure this is present in the target env/Helm/secret setup before rollout.

From my side the original change requests are resolved.

PR #392, opened 2026-07-11
feat: wire invite two-factor gate to real OTP lifecycle

Storypapst:
## Why

The account-invite two-factor gate (`TwoFactorGateStatus`) was scaffolding: counsellor invites default to `PENDING_SETUP`, but nothing ever transitioned the status — no path to `ACTIVE`, and `waiveTwoFactor` had no REST endpoint and never persisted. The gate could block counsellors (`BLOCKED_TWO_FACTOR`) with no way to satisfy it.

This lands together with [ORISO-Helm#55](https://github.com/OpenResilienceInitiative/ORISO-Helm/pull/55), which makes the underlying 2FA endpoints actually work (otp-config SPI in the Keycloak image).

## What

- `PENDING_SETUP → ACTIVE` when the user activates an OTP credential (app **and** email flow), `ACTIVE → PENDING_SETUP` when the credential is deleted. Waived invites stay untouched.
- New endpoint `POST /useradmin/account-invites/{inviteId}/waive-two-factor` (same admin authorities as the other invite admin routes; reason required) persisting `twoFactorWaivedBy/At/Reason`.
- `waiveTwoFactor` now persists; id-based overload for the controller.
- New repository finder `findAllByAcceptedByUserIdAndTwoFactorStatus`.

## Tests

- `AccountInviteServiceTest`: 4 new tests (transitions both ways, blank-user/no-match no-ops, waive persistence) — 50/50 green.
- `UserTwoFactorAuthControllerDelegateTest`: extended for the new wiring — 16/16 green.
- Note: `UserController2faE2EIT` fails on pre-dev with 28×401 **before this change** — the IT auth harness issue also hits other E2EITs (e.g. `UserControllerSessionE2EIT` 19/29 red); tracked by the test-quality audit, not addressed here.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/AccountInviteController.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/AccountInviteController.java
@@ -10,6 +10,7 @@ import de.caritas.cob.userservice.api.service.accountinvite.AccountInviteService
 import de.caritas.cob.userservice.api.service.accountinvite.AccountInviteService.CreateAccountInviteCommand;
 import de.caritas.cob.userservice.api.service.accountinvite.AccountInviteService.InviteSendResult;
 import de.caritas.cob.userservice.api.service.accountinvite.AccountInviteService.SendInviteCommand;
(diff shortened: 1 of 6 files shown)


Storypapst:
Part of epic #393 (WP-6).

PR #416, opened 2026-07-12
Switch tracing/metrics export from dead Zipkin to OTLP for SigNoz

Storypapst:
## Summary

Part 1/4 of OBS-P2 (observability two-track plan, `0 - Docs/PLAN-observability-two-track-2026-07-11.md`). SigNoz is live on Pre-Dev with a working log pipeline; this closes the traces + app-metrics gap for UserService.

- Removes the `opentelemetry-exporter-zipkin` dependency and the `spring.zipkin.*` / `spring.sleuth.*` config block. No Zipkin collector runs anywhere - this was dead code producing the same connection-refused noise as `ORISO-AGENCYSERVICE-1` (`Connect to http://localhost:9411 ... Connection refused`).
- Adds `org.springframework.boot:spring-boot-starter-opentelemetry` (the official Spring Boot 4 starter) and wires `management.otlp.tracing.endpoint` / `management.otlp.metrics.export.url` as env-driven properties (`MANAGEMENT_OTLP_TRACING_ENDPOINT` / `MANAGEMENT_OTLP_METRICS_EXPORT_URL`), mirroring the existing `sentry.dsn=${SENTRY_DSN:}` pattern. Empty by default until ORISO-Helm sets the env vars per environment.
- Pins `com.google.protobuf:protobuf-java` to `4.32.0` (see "Non-obvious findings" below).
- Adds `traceId`/`spanId` to the JSON log encoder's MDC output so SigNoz can correlate a log line to its trace.
- Fixes two pre-existing UserService issues while in the area:
  - `logback-spring.xml`: the `testing` profile's plain-text pattern read MDC key `X-Correlation-Id`, which `CorrelationIdFilter` never sets (it sets `CID`). Now reads `%X{CID}`, consistent with the `!testing` JSON profile.
  - `application-prod.properties`: dropped leftover `logging.level.org.springframework.web=DEBUG` and `...mvc.method.annotation=DEBUG` (still present in dev/staging/testing, only prod was the leftover).

## Non-obvious findings (verified empirically this session)

Spring Boot 4 split what used to be a monolithic `spring-boot-actuator-autoconfigure` into many per-feature autoconfigure modules. Two things had to be verified rather than assumed:

1. **A naive dependency swap silently does nothing.** Just swapping `opentelemetry-exporter-zipkin` → `opentelemetry-exporter-otlp` + `micrometer-registry-otlp` alongside the existing `micrometer-tracing-bridge-otel` leaves the actual `OpenTelemetry` SDK bean unregistered (that autoconfiguration now lives in `org.springframework.boot:spring-boot-micrometer-tracing-opentelemetry`, which isn't pulled in transitively by the starter or by the micrometer/otel jars alone - confirmed the **previous** Zipkin setup never had it on the classpath either, i.e. Zipkin tracing was already fully inert here). Using the official `spring-boot-starter-opentelemetry` starter bundles everything needed correctly.
2. **Empty-endpoint no-op required a protobuf-java version bump.** With the starter in place but `management.otlp.*` endpoints empty, context startup failed with `NoClassDefFoundError: com/google/protobuf/RuntimeVersion$RuntimeDomain`. Root cause: `io.opentelemetry:opentelemetry-proto` needs `protobuf-java 4.32.0+` generated code, but the `firebase-admin` → `grpc`/`google-cloud-core-grpc` dependency chain pulls a much older `protobuf-java 3.17.3`, and Maven's nearest-wins mediation picked that one. Pinned `protobuf-java` to `4.32.0` as a direct dependency to force resolution. After this fix, verified clean app-context boot with both OTLP endpoints empty - **no boolean enable/disable toggle needed**, the plain env-driven empty-string pattern (same as `sentry.dsn`) no-ops correctly.

## For ORISO-Helm (cross-service consistency)

Use the env-driven empty-string pattern, not `management.otlp.*.export.enabled` booleans:
```
MANAGEMENT_OTLP_TRACING_ENDPOINT=
MANAGEMENT_OTLP_METRICS_EXPORT_URL=
```
Other services on Spring Boot 4 with the same `firebase-admin`/old-grpc dependency chain will likely need the same `protobuf-java` pin to avoid the same `NoClassDefFoundError`.

## Test plan

- [x] `mvn compile` (JDK21) - green, spotless check passes
- [x] `mvn spotless:apply` (JDK21) - clean, no additional changes
- [x] Full unit test suite (JDK21): 3395 tests, 0 failures, 0 errors
- [x] `UserServiceApplicationIT` (`@SpringBootTest`, boots full context) under JDK21 and JDK25, both green, with both OTLP endpoints empty - confirmed no export-related errors/exceptions in logs
- [x] Confirmed root-caused two real startup failures during the empty-endpoint check (missing OTel SDK bean; protobuf-java version conflict) and fixed both rather than assuming they'd no-op

Note: `mvn test` on this machine's default JDK25 fails ~2685 tests with `Mockito cannot mock this class` (ByteBuddy/JDK25 incompatibility) - confirmed this is pre-existing on unmodified `origin/pre-dev` too, unrelated to this change. All verification above ran under JDK21 where the full suite is green.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -138,13 +138,33 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-core</artifactId>
 		</dependency>
+		<!-- Traces + JVM/app metrics for SigNoz via OTLP. Replaces the previous
+			micrometer-tracing-bridge-otel + opentelemetry-exporter-zipkin combo:
+			no Zipkin collector runs anywhere, it only produced dead
+			ResourceAccessException/Connection-refused noise against localhost:9411
+			(ORISO-AGENCYSERVICE-1). This single Spring Boot 4 starter bundles the
+			OpenTelemetry SDK bean, the OTel tracing bridge, OTLP trace export and
+			OTLP metrics export - Boot 4 split these into separate autoconfigure
+			modules, and assembling them by hand (tracing-bridge-otel + the raw
+			OTLP exporter jars alone) leaves the actual OpenTelemetry SDK bean
+			missing, which fails context startup. Verified empirically: with this
+			starter and empty management.otlp.* endpoints, the app boots cleanly. -->
 		<dependency>
-			<groupId>io.micrometer</groupId>
-			<artifactId>micrometer-tracing-bridge-otel</artifactId>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-opentelemetry</artifactId>
 		</dependency>
+		<!-- io.opentelemetry:opentelemetry-proto requires protobuf-java 4.32.0+
+			generated code (io.grpc + com.google.protobuf.RuntimeVersion checks).
+			The firebase-admin -> grpc/google-cloud-core-grpc chain below pulls the
+			much older protobuf-java 3.17.3, and Maven's nearest-wins mediation
+			picked that older one, which crashed context startup with
+			NoClassDefFoundError: com/google/protobuf/RuntimeVersion$RuntimeDomain
+			(verified empirically). Declaring it directly here wins over the
+			transitive version regardless of tree depth. -->
 		<dependency>
-			<groupId>io.opentelemetry</groupId>
-			<artifactId>opentelemetry-exporter-zipkin</artifactId>
+			<groupId>com.google.protobuf</groupId>
+			<artifactId>protobuf-java</artifactId>
+			<version>4.32.0</version>
 		</dependency>
 
 		<!-- Spring actuator  -->
--- a/src/main/resources/application-prod.properties
+++ b/src/main/resources/application-prod.properties
@@ -1,9 +1,7 @@
 # ---------------- Logging ----------------
 server.port=8082
 logging.level.root=INFO
-logging.level.org.springframework.web=DEBUG
 logging.level.org.hibernate=ERROR
-logging.level.org.springframework.web.servlet.mvc.method.annotation=DEBUG
 # Keep RestTemplate at INFO: its DEBUG body logging prints outbound request
 # payloads (e.g. Matrix service-account passwords) in clear text.
 logging.level.org.springframework.web.client.RestTemplate=INFO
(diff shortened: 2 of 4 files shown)


Storypapst:
## Fix pushed: OTLP property names corrected

Independently verified against the actual `spring-configuration-metadata.json` inside `spring-boot-micrometer-tracing-opentelemetry-4.0.1.jar` (not docs): `management.otlp.tracing.endpoint` carries `"deprecated": true, "level": "error", "since": "4.0.0"` with replacement `management.opentelemetry.tracing.export.otlp.endpoint` - the property this PR originally used was dead in Boot 4.0.1, so tracing was exporting nothing regardless of value.

**Corrected property set** (now in `application.properties`):
```properties
management.opentelemetry.tracing.export.otlp.endpoint=${MANAGEMENT_OTLP_TRACING_ENDPOINT:}
management.tracing.export.otlp.enabled=${MANAGEMENT_OTLP_TRACING_ENABLED:false}
management.otlp.metrics.export.url=${MANAGEMENT_OTLP_METRICS_EXPORT_URL:}
management.otlp.metrics.export.enabled=${MANAGEMENT_OTLP_METRICS_EXPORT_ENABLED:false}
```

Also confirmed empirically (not assumed) that an empty endpoint/url alone is NOT a safe no-op for either exporter - Spring's `OnPropertyCondition` treats "property present with any value, including empty string" as a match. Each exporter therefore needs its own explicit `*_ENABLED` flag, defaulting to `false`.

**Verification**, running the real app (not just a short `@SpringBootTest`) for 2+ minutes past the metrics registry's 1-minute publish cycle:
- Both `*_ENABLED` unset (default `false`): zero otlp/export-related log lines over 2+ minutes uptime. Spring's `--debug` condition-evaluation report confirms neither `OtlpTracingConfigurations.Exporters` nor `OtlpMetricsExportAutoConfiguration` create their exporter beans in this state.
- Fake endpoint (`http://127.0.0.1:19999`) + both `*_ENABLED=true`: condition report confirms `OtlpTracingConfigurations.Exporters#otlpHttpSpanExporter` is created; the metrics `OtlpMeterRegistry` made a real HTTP POST to the fake endpoint and logged a genuine `java.net.ConnectException: Connection refused`, both on its 1-minute timer and at graceful shutdown.

Full unit suite (3395 tests) and the application-boot IT stay green under JDK21.

**For ORISO-Helm:** use this exact 4-property set (2 endpoints/urls + 2 enabled booleans, all defaulting to empty/`false`) for consistency across all 4 services.

Issue #413, opened 2026-07-12
[PreDev security] Anonymous invite-link fallback resolves a cross-tenant agency


Storypapst:
## Problem (cross-tenant isolation defect found by Gate B live chat)

An External Inbound **Live Chat** invite link scoped to **tenant 83 / topic 11** with no explicit `agencyId` enrolls the anonymous asker into **Agency 237, which belongs to tenant 1** ("Caritas Agency"). The created session lands on **tenant 1**, not tenant 83.

## Root cause

`AgencyInviteLinkService.resolveAgencyIdForRegistration` (UserService) falls back, when the link has no `agencyId`, to `agencyService.getAgenciesByConsultingType(consultingTypeId)` and takes `agencies.get(0)`. That AgencyService lookup is **not tenant-scoped**: for `consultingType=1` it returns agencies across tenants (verified: `GET /service/agencies?postcode=10115&topicId=11&consultingType=1` returns agencies of tenants 1, 82, 83 regardless of the `tenantId` header), and the first element is agency 237 (tenant 1).

Even though `redeem()` sets `TenantContext` to the link's tenant (83) before the call, the downstream AgencyService endpoint ignores the tenant header, so the fallback silently crosses tenants. The FE (`InviteLink.tsx`) then registers the anonymous asker with `agencyId: 237` as the tenant-resolving header, so `HttpTenantFilter` resolves tenant **1** and the session is created there.

Consequences:
- A tenant-83 live-chat link routes the asker into a different Träger's agency (tenant-1 "Caritas Agency") — a cross-tenant leak.
- The asker session is not visible in the intended tenant-83 consultant's live-chat queue, so Gate B live chat cannot complete.

## Reproduction

1. Tenant admin of tenant 83 creates `POST /service/useradmin/invitelinks` with `linkKind=EXTERNAL_INBOUND, chatType=LIVE_CHAT, anonymity=FULL, topicId=11` (no agencyId). → link 268, `tenantId=83`.
2. `POST /service/users/invitelinks/{token}/redeem` → returns `agencyId: 237, tenantId: 83, sessionId: null`.
3. Agency 237 is tenant 1. Registration via that agency header creates a tenant-1 session.

## Acceptance criteria

- The no-`agencyId` fallback resolves an agency **belonging to the link's tenant** (and, when the link is topic-bound, one that serves that topic — e.g. agency 280 for tenant 83 / topic 11).
[...]

PR #421, opened 2026-07-12
fix: Matrix-migrate the anonymous live-chat path (RC-disabled)

Storypapst:
## Why

Follow-up to #417. Deploying the anonymous topic-queue flow to PreDev (Gate B of E2E run `e2e-20260711-1357`) surfaced that the **entire anonymous live-chat backend path was never Matrix-migrated** — it hard-required Rocket.Chat and broke under `rocket-chat.enabled=false` (ADR-004). Each fix below is a live-RED → TDD repair.

## What

1. **`AnonymousUserCreatorService` — Matrix-only account.** Gated the Rocket.Chat user create/login behind `rocket-chat.enabled`; when disabled, carry Matrix-migration dummy RC credentials. Also provision a Matrix account for the anonymous asker (the path called only `updateIdentityAndCreateAccount`, which does not create a Matrix user, so the first enquiry failed "has no Matrix account"). Extracted the Matrix-user provisioning from the registered path into a reusable **`CreateUserFacade.ensureMatrixUser`** (registered-path refactor is behavior-preserving — 26 tests green) and call it from the anonymous path in Matrix-only mode.
2. **`AnonymousConversationCreatorService` — no room while waiting.** It always created a Rocket.Chat room. A waiting live-chat session has no agency and needs no room yet: the Matrix room is created on **accept** (`AssignEnquiryFacade.assignAnonymousEnquiry`, which reuses or creates a room), which is also when the agency/tenant binding + that agency's DSE apply. Gated room creation behind `rocket-chat.enabled`.
3. **`LiveEventNotificationService` — best-effort delivery.** Only the declared `ApiException` was caught; a transport failure to the live service (host unreachable → `UnresolvedAddressException`/`ConnectException` wrapped in a `RuntimeException` by the java.net.http client) escaped and broke the enquiry/session flow. Broadened the catch — live events are best-effort; consultants reconcile via the next queue poll.

## Tests

- `AnonymousUserCreatorServiceTest` (RC-disabled skip path + existing RC-enabled cases), `AnonymousConversationCreatorServiceTest` (RC-disabled room-skip), `LiveEventNotificationServiceTest` (RuntimeException swallowed), `CreateUserFacadeTest` (26, behavior-preserving).
- Full Java 21 unit gate: 3,431 tests, 0 failures, 0 errors, 7 skipped; Spotless applied.

## Separate config finding (ORISO-Helm, not code)

PreDev `userservice-configmap-env` had `LIVE_SERVICE_API_URL=http://liveservice.caritas:8080` but the `liveservice` Service listens on **8086** (short host also didn't resolve from the pod). Hot-patched the live ConfigMap to `http://liveservice.caritas.svc.cluster.local:8086` for the E2E run; needs a durable ORISO-Helm fix.

Refs #413. PreDev only; no merge or deployment to dev.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/conversation/service/AnonymousConversationCreatorService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/conversation/service/AnonymousConversationCreatorService.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Value;
 import org.springframework.stereotype.Service;
 
 /** Service to create anonymous user conversations (sessions). */
@@ -38,6 +39,10 @@ public class AnonymousConversationCreatorService {
   private final @NonNull ConsultantAgencyService consultantAgencyService;
   private final @NonNull LiveEventNotificationService liveEventNotificationService;
   private final @NonNull TopicConsultantRoutingService topicConsultantRoutingService;
+
+  /** ADR-004: with Rocket.Chat disabled no Rocket.Chat room is created for the waiting session. */
+  @Value("${rocket-chat.enabled:false}")
+  private boolean rocketChatEnabled;
 
   /**
    * Creates a new anonymous conversation session with the corresponding Rocket.Chat room.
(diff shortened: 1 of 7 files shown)


Storypapst:
Superseded by #420 (create anonymous live chat in Matrix-only mode), which landed in parallel and covers the same RC-gating + room-skip. The asker Matrix account is created lazily on accept (`AssignEnquiryFacade.ensureUserMatrixAccount`), so the extra `ensureMatrixUser`-at-creation here is redundant. The one unique, still-useful change — making `LiveEventNotificationService` delivery best-effort against transport failures (RuntimeException, not just ApiException) — will be re-submitted as a small standalone hardening PR. Closing to avoid duplication. Refs #413/#419.

PR #474, opened 2026-07-18
fix(notifications): make message notifications work in E2EE Matrix rooms

kiodreambau:
## Why

A literal 2-user E2E click-through on pre-dev (broker-driven Playwright: asker sends, consultant watches live) showed that **no message notification is ever created** for 1:1 session rooms, even though the writer/live-event/feed pipeline (#443/#474) is deployed and healthy. Three independent structural gaps:

1. **`m.room.encrypted` was ignored.** All session rooms are E2EE, so every real message arrives as `m.room.encrypted` and fell through the listener's event-type switch into `default: ignore`.
2. **NPE on encrypted events.** `buildPrivacyEnvelope` used a null-hostile `Set.of(...).contains(msgtype)` — encrypted events have no `msgtype`.
3. **The admin was in no room.** The listener /sync loop authenticates as the technical Matrix admin, but `rooms.join` in /sync only contains rooms that user has *joined* — and nothing ever added the admin to session rooms. Verified live: after manually inviting+joining the admin into a test room, the loop immediately logged `Processing events for registered room`.

## What

- Route `m.room.encrypted` through the message handler (metadata-only: sender + event id are cleartext; content preview mode is NONE anyway — no plaintext crosses the boundary, FE-H01 intact).
- Null-guard `msgtype` in `buildPrivacyEnvelope`.
- New `MatrixSynapseService.ensureAdminInRoom(roomId, memberMatrixUserId)`: direct join, falling back to member-impersonated invite + join for invite-only rooms. Called
  - at room creation (`AssignEnquiryFacade.createNewMatrixRoom`, `DirectSessionMatrixRoomService`) → all new rooms work out of the box,
  - as a best-effort lazy heal in `POST /matrix/sync/register/{sessionId}` → existing rooms heal when opened (FE companion PR calls this endpoint).

## Tests

TDD (red first for each): `MatrixEventListenerServiceTest` (+2), `MatrixSynapseServiceTest` (+3), `MatrixSyncControllerTest` (+2), `AssignEnquiryFacadeTest` (+1), `DirectSessionMatrixRoomServiceTest` (+1). Full unit suite: **3628 tests, 0 failures**.

## Ops notes (pre-dev)

- Secret vs. ConfigMap drift: the effective Matrix admin username comes from `oriso-platform-userservice-secrets` (`matrixadm`), while the ConfigMap still says `caritas_admin` — worth aligning.
- The session-room registry and admin-membership healing are lazy; no data migration needed.

Refs OpenResilienceInitiative/ORISO-Frontend#436 (notifications epic)

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixSynapseService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/matrix/MatrixSynapseService.java
@@ -884,6 +884,56 @@ public class MatrixSynapseService {
   }
 
   /**
+   * Ensures the technical Matrix admin is a member of the given room. The event-listener /sync loop
+   * runs with the admin account and only receives events for rooms the admin has joined — without
+   * this membership, message notifications for the room can never fire.
+   *
+   * <p>Invite-only session rooms reject a direct admin join, so an existing member is impersonated
+   * (Synapse admin login-as-user) to invite the admin first.
+   *
+   * @param roomId the Matrix room ID
+   * @param memberMatrixUserId full Matrix ID of a current room member used for the invite fallback
+   * @return true when the admin is (now) a member of the room
+   */
+  public boolean ensureAdminInRoom(String roomId, String memberMatrixUserId) {
+    if (roomId == null || roomId.isBlank()) {
+      return false;
+    }
+    String adminToken = getAdminToken();
+    if (adminToken == null) {
+      return false;
+    }
+
+    // Direct join succeeds for public rooms or when the admin is already invited/joined.
+    if (joinRoom(roomId, adminToken)) {
+      return true;
+    }
+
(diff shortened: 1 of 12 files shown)


kiodreambau:
Follow-up commit `a2368e6` — live verification on pre-dev (plaintext probe into the healed room) surfaced a **fourth gap**: the async notification task died with `IllegalStateException: No thread-bound request found`, because `LiveEventNotificationService.notInitiatingUser` dereferences the request-scoped `AuthenticatedUser` (fatal on the sync-loop thread), and the shared try/catch let that failure also swallow the `event_notification` row.

Fixed both: no request context → nobody to exclude (event still goes out), and live-send vs. notification-row creation are now separate failure domains. TDD (+2 tests), full suite 3630 green.

_🤖 [Claude Code](https://claude.com/claude-code)_

Issue #495, opened 2026-07-18
[SUPPORT-01] Live-Handshake backend: session entity, fresh re-auth, 5-min TTL, audit + 12-month retention


kiodreambau:
## Why
ADR-018 §1/§3: the reusable two-person confirmation primitive with full audit.

## What to build
- Handshake session entity (initiator, counterpart, purpose enum, TTL ≈5 min, state machine) with single-use semantics.
- Endpoints: initiate (verifies initiator's fresh password+OTP via Keycloak), confirm (verifies counterpart's fresh password), expire (lapse → audit entry 'session was not established').
- Audit entries for every transition; scheduled deletion after 12 months.
- Authority model: purposes declare the required role pair (support-access, recovery, identity-grant).

## Acceptance
- [ ] No handshake completes without two distinct authenticated humans with fresh credentials.
- [ ] Lapsed handshakes leave exactly one audit entry and no other state.
- [ ] Retention job deletes audit entries older than 12 months.

Part of epic OpenResilienceInitiative/ORISO-UserService#493 (ADR-018)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Storypapst:
Verified as delivered: PR #502 (merged to pre-dev) delivers the full live-handshake primitive from ADR-018: HandshakeSession/HandshakeAuditEvent entities with Liquibase migrations, a HandshakeService enforcing two distinct authenticated people with fresh password+OTP (initiator) and fresh password (counterpart), a ~5-minute TTL with a scheduled lapse sweep that leaves exactly one audit entry, and a scheduled 12-month audit-retention purge — matching all three acceptance criteria, confirmed directly in the merged HandshakeService.java source.

Moving to **QA external Dev** for external review. cc @BjoernLudwig

Issue #493, opened 2026-07-18
[SUPPORT-01] EPIC: Global Support Admin, Live-Handshake primitive, Temporary Support Access (ADR-018)


kiodreambau:
Affected repos: ORISO-UserService, ORISO-Admin, ORISO-Frontend (ORISO-ElementCall pointer tracked separately)

## Why

ADR-018 (decided 2026-07-18, all grill security gates closed): support help happens in a **fresh encrypted 1:1 Matrix room** created only after a **live handshake** — Support Admin re-authenticates with password+OTP, the affected Berater*in confirms with their password in an in-app popup (≈5 min window; lapse → audit entry only). Scope is exactly that chat + its audio/video call; hard 4-hour limit, rooms never reused ("komplett von vorne"); no emails, no break-glass. The handshake module is a reusable primitive that also carries four-eyes account recovery and identity grants. Platform Admin remains unchanged; Global Support Admin is an additive, separate identity with its own table.

Decision trail: `0 - Docs/plans/2026-07-18-multi-role-users-temporary-support-access-grill-report.md`; ADR: `0 - Docs/ADRs/ADR-018-global-support-admin-live-handshake-temporary-support-access.md` (workspace).

## What to build (sub-issues)

- [ ] US: Live-Handshake backend + audit + 12-month retention
- [ ] US: Global Support Admin identity (role, admin type, mandatory-2FA gate)
- [ ] US: Support-room lifecycle (dual-confirm create, 4h kick+archive)
- [ ] Admin: GSA table, support-button column, handshake surfaces, audit views
- [ ] Frontend: Berater*in confirm popup + support chat/call surface
- [ ] US: Four-eyes recovery flows (GSA and Berater*in)
- [ ] US+Admin: identity grants re-gated onto the handshake (tightens slice-A grant)

Suggested build order: handshake backend → GSA identity → room lifecycle → Admin/Frontend surfaces → recovery → grant hardening. E2E proof per visible workflow required before promotion.

## Acceptance

- [ ] A support session can only ever come into existence through the dual live handshake; no email, no unilateral path.
- [ ] The room exposes zero pre-existing data and dies hard at 4 hours (or earlier termination) — kicked + archived, never reused.
- [ ] Audit: Platform Admin sees all, the affected Beratungsstelle sees its own entries; auto-deletion after 12 months.
- [ ] No recovery or identity grant executes with fewer than two authorized humans.

[...]

PR #502, opened 2026-07-19
feat(handshake): live-handshake core primitive — sessions, fresh re-auth, TTL, audit, retention

kiodreambau:
## Problem
ADR-018 needs its core primitive: two people confirming a privileged action live with fresh credentials before anything executes — the base for support-room creation, four-eyes recovery and identity grants.

## Changes
- `handshake_session` + `handshake_audit_event` tables (Liquibase 0072), entities, repositories.
- `HandshakeService`: initiate (fresh password+OTP via new `KeycloakAuthClient.verifyWithOtp`, per-purpose role check, two-distinct-people rule, ~5-min TTL), confirm (counterpart password via `verifyIgnoringOtp`, purpose role check, single-use), pending lookup for the popup, scheduled lapse sweep ("session was not established" audit entry), scheduled 12-month audit purge.
- `HandshakePurpose` declares role pairs for SUPPORT_ACCESS / RECOVERY_SUPPORT_ADMIN / RECOVERY_CONSULTANT / IDENTITY_GRANT; `HandshakeCompletionHandler` beans run inside the confirm transaction (consumers land in US#497/498/499).
- REST: `POST /users/handshakes`, `POST /users/handshakes/{id}/confirm`, `GET /users/handshakes/pending` (identities strictly from the token) + SecurityConfig matcher.

## Verification
- TDD: 21 new unit tests (service state machine, role gates, TTL, auth client, controller); full suite green (3709 tests).

Closes OpenResilienceInitiative/ORISO-UserService#495
Refs OpenResilienceInitiative/ORISO-UserService#493

## Summary by CodeRabbit

* **New Features**
  * Added secure live handshakes for initiating, confirming, and viewing pending privileged actions.
  * Added temporary support rooms with active-room listing, termination, expiration, and lifecycle management.
  * Added support-admin search with pagination.
  * Added global support-admin role and clearer “Support Admin” labeling.
  * Added password-and-OTP verification for supported authentication flows.

* **Security**
  * Added stronger authorization, tenant isolation, credential validation, confirmation limits, and audit tracking for handshakes.


shazia-k:
@kiodreambau here are the results of PR review
The branch does implement the ADR-018 core primitive correctly: two distinct people, fresh credentials (password+OTP for initiator, password for counterpart), a TTL window, single-use sessions, transactional action dispatch, and an audit trail with retention. The skeleton is sound.
The most important fixes before this is used as a base for support-room creation or identity grants are #2 (tenant isolation on confirm), #3 (brute-force guard on the confirm endpoint), and #6 (optimistic lock to prevent double-confirmation). Items #1 and #4 are correctness risks; #5 and #7 are hardening/ops issues.

shazia-k (commented):
@kiodreambau 
Here are the concerns to fix 

🔴 Must-fix before merge (2 items)
TOCTOU race on confirm() — two concurrent requests can both see status == PENDING before either updates it. Add @Version long version to HandshakeSession (optimistic locking). On MySQL READ_COMMITTED, @Transactional alone does not prevent this.
Fragile Keycloak OTP-bypass detection — verifyIgnoringOtp returns true if the error body contains "Missing totp". That string is Keycloak-internal, not a stable contract. A Keycloak upgrade could silently break or bypass password verification for counterparts.
 
🟡 Should-fix (4 items)
No @Valid / size constraints on request DTOs — password, otp, counterpartId accept unlimited input that is passed to Keycloak.
Role-check failure produces no audit entry — CONFIRM_REJECTED is only written for password failures; a role mismatch is silent in the audit trail.
LocalDateTime.now() is timezone-unaware — container JVM ≠ DB timezone breaks TTL comparisons. Use LocalDateTime.now(ZoneOffset.UTC).
sweepExpired() has no upper bound — a large backlog returns all rows into memory at once; paginate.

Storypapst:
Deep readiness review: this PR should remain blocked even though its checks are green.\n\nThe core ADR-018 skeleton is present, but the current confirmation path has three concrete security and consistency gaps:\n\n1. Tenant isolation: the session stores the initiator tenant, but confirm only checks counterpart ID and role. It never verifies that the authenticated counterpart belongs to the session tenant. The pending query is also keyed only by counterpart ID. Both confirm and pending lookup need tenant-scoped repository methods or an explicit same-tenant/platform-admin policy, with negative cross-tenant tests.\n2. Brute-force protection: every wrong counterpart password reaches Keycloak and writes an audit event, but there is no per-session/per-actor attempt limit, cooldown, or terminal lock state. A durable attempt counter and threshold should be enforced transactionally, in addition to upstream Keycloak protections.\n3. Double confirmation: findById followed by an ordinary save permits two concurrent requests to observe PENDING and both execute completion handlers. Add optimistic locking with a version column or a conditional PENDING-to-CONFIRMED update and make handler execution depend on exactly one successful transition.\n\nThe earlier review summary also refers to findings 1 through 7, but their descriptions are not present in any inline thread or visible comment. The missing details need to be restored before a complete fix can be verified. Because this path authorizes privileged actions, I do not recommend guessing at the remaining requirements or applying a partial autofix. Please add the seven concrete findings; then the above three should be implemented with concurrency and cross-tenant tests before approval.

kiodreambau:
@shazia-k @Storypapst thanks — all findings are addressed in 3375f8b (and propagated up the stack to #504/#506):

**Must-fix**
1. **TOCTOU / double confirmation** → `@Version` optimistic lock on `handshake_session` (column in the 0072 changeset, which is unmerged so no migration split). The PENDING→CONFIRMED transition is saved **before** handler dispatch; a concurrent loser gets `OptimisticLockingFailureException` → 400, and completion handlers never run for it (test: `confirm_Should_RejectWithoutHandlers_When_ConcurrentConfirmationWinsTheRace`).
2. **Fragile "Missing totp" detection** → `verifyIgnoringOtp` now parses the JSON body and accepts only the exact `error_description: "Missing totp"` contract of the **vendored** otp-config SPI (ADR-013 — the string is version-pinned under ORISO control, documented in the code). Negative test: a body merely *containing* the phrase no longer passes.

**Tenant isolation** → per-purpose policy: `SUPPORT_ACCESS` is cross-tenant **by design** (platform support entering a lower scope is the entire feature, ADR-018 §2); every other purpose requires the confirming counterpart to be in the session tenant or platform-scoped (tenant 0). Negative cross-tenant test for `RECOVERY_CONSULTANT`, positive cross-tenant test for `SUPPORT_ACCESS`. The pending lookup remains token-scoped (`counterpartId` from JWT only).

**Brute force** → durable `confirm_attempts` counter, incremented transactionally on every failed password; at the limit (default 5, `handshake.max-confirm-attempts`) the session locks terminally with a `CONFIRM_LOCKED` audit entry.

**Should-fix** → bean-validation `@NotBlank`/`@Size` on both DTOs with `@Valid`; role-mismatch confirms now write `CONFIRM_REJECTED` audit entries; all timestamps via `CustomLocalDateTime.nowInUtc()`; `sweepExpired` processes bounded batches (default 200, `handshake.sweep-batch-size`). The same UTC + bounded-sweep hardening is applied to the support-room lifecycle in #506.

Full unit suite green (3716 on this branch, 3730 on top of the stack).

shazia-k (approved):

kiodreambau:
Rebased onto current `pre-dev` — conflicts resolved, PR is mergeable again.

**Conflict:** `pre-dev` introduced `0072_user_do_not_disturb` while this branch carried an unmerged `0072_handshake` + `0073_support_room`. Since neither has ever been merged or deployed, mine are **renumbered to `0073_handshake` / `0074_support_room`** (directory, file **and** changeset id) so the master changelog stays strictly ordered and collision-free. `SecurityConfig` and `KeycloakServiceTest` auto-merged.

**Adopted from parallel work on `pre-dev`** (thanks @shazia-k):
- `adadd471` (*stop deadlocking platform admins behind an unsatisfiable 2FA gate*) applies to the GSA gate here too: if the OTP role policy denies OTP for the support-admin role, enrollment is impossible and the gate can never be satisfied. It now checks `identityClientConfig.isOtpAllowed(roles)` first — **still fails closed**, but reports a deployment configuration error (`identity.otp-allowed-*`) instead of asking the admin to complete an impossible enrollment. New test: `initiate_Should_ReportConfigurationError_When_OtpPolicyDeniesOtpForSupportAdmins`.
- `fc2be917` / `a5dfd306` (OTP username encoding + admin-token refresh): the OTP lookup here already follows the canonical `encodeUsername` pattern of `UserAccountControllerDelegate`, and `verifyIgnoringOtp` now carries the `decodeUsername` from `pre-dev` — no divergence.
- `e0d7551a` restored the public-slug changelog guard, so my earlier #489 is fully superseded on `pre-dev`; nothing left to do there.

Note that #504 and #506 merged into this branch, so **this PR now carries the whole stack**: handshake primitive + GSA identity + support-room lifecycle (35 files vs `pre-dev`).

Full unit suite green after the merge: **3838 tests, 0 failures**.

kiodreambau:
Rebased onto current `pre-dev` again and fixed a real defect the gate caught.

**Changeset renumbering (second time):** `pre-dev`'s Rocket.Chat removal claimed `0073`–`0075`, so the still-unmerged handshake/support changesets moved once more — now **`0076_handshake` / `0077_support_room`** (directory, file and changeSet id). Because these numbers keep colliding while this PR waits, please re-check the tail of `userservice-master.xml` right before merging.

**Defect found by `mariadb-contract`:** the previous conflict resolution dropped the opening `<include>` tag before `0073_remove_rocket_chat_user_ids`, leaving stray character data inside `databaseChangeLog`. Liquibase then refused to parse the master changelog — i.e. **every fresh-database boot would have failed**. Fixed, and verified locally against a clean MariaDB (`DatabaseChangelogDriftIT` + `AdminStatisticsRepositoryMariaDbIT` both green). Good catch by that gate; the unit suite alone would never have seen it.

No functional change to the handshake/support code in these two commits.

Issue #507, opened 2026-07-19
Fix own-email rejection during consultant 2FA setup


Storypapst:
## Why

Consultants created in Admin have a raw username in Keycloak and an encoded username in UserService. During email 2FA setup, the ownership check compares only identical representations, so the consultant’s existing email is rejected as already in use.

## What to build

- Normalize the authenticated username before comparing it with the Keycloak email owner.
- Preserve legacy encoded-username compatibility and rejection of emails owned by another user.
- Add focused regression coverage for raw and encoded Keycloak usernames.

## Acceptance

- [x] A consultant created in Admin can start email 2FA with the same email used at creation.
- [x] An email owned by another account remains rejected.
- [x] Focused unit and controller tests pass.
- [ ] Pre-Dev behavior is verified with the Dreambau Test Access account.

Parent: OpenResilienceInitiative/ORISO-UserService#393

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Storypapst:
Implementation update: PR #508 was approved and merged to pre-dev as 95b9e66; all PR and feature-branch checks passed, and the fixed flow was proven on desktop/mobile with the Dreambau Test Access consultant. Runtime gate remains: GitHub did not create the expected CI - Main push run for merge commit 95b9e66, and that workflow has no workflow_dispatch trigger. PreDev was restored to the prior regular digest after hot proof, so regular image publication, deploy, and one focused post-deploy retest are still required before moving this issue to Done.

Storypapst:
Release recovery work package: native sub-issue #509 -> PR #510. The PR adds only a manual dispatch trigger while preserving the existing push triggers. After human review and merge, the dispatch will publish the current pre-dev head for regular deployment and final Dreambau-account retest.

kiodreambau:
## Verification run 2026-08-19 — the code fix landed and is deployed

Re-verified against live GitHub and the running Pre-Dev image, not against this
issue's own claims. Summary: **the defect is fixed in code and the fix is in the
image currently serving Pre-Dev.** What is still genuinely open is the
confirmatory retest, which is a much smaller thing than this issue's open state
suggests.

### Evidence per acceptance criterion

| Acceptance criterion | State | Evidence |
|---|---|---|
[...]

PR #508, opened 2026-07-19
Fix own-email rejection during consultant 2FA setup

Storypapst:
## Why

Admin-created consultants have a raw username in Keycloak and an encoded username in UserService. The email ownership check compared only identical representations, so the consultant’s existing email was rejected as already in use during email 2FA setup.

## What changed

- Decode the authenticated username before comparing it with Keycloak’s decoded owner username.
- Preserve the legacy encoded-username path and rejection of another account’s email.
- Add focused manager coverage and make the controller E2E fixture reflect the current raw Keycloak username model.

## Verification

- Red: raw Keycloak username regression returned `false` before the fix.
- Green: `IdentityManagerTest` + `UserTwoFactorAuthControllerDelegateTest`: 20 passed.
- Full unit suite: 3,713 passed, 0 failed, 7 skipped.
- Spotless check: passed under Java 21.
- The legacy `UserController2faE2EIT` harness currently returns 401 for every protected scenario locally, including unchanged tests, before application logic; this is separate from the product behavior.

Closes OpenResilienceInitiative/ORISO-UserService#507

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email ownership checks when usernames are encoded or decoded, ensuring users can correctly retain their own email address.
  * Updated the two-factor authentication email flow for accounts using raw Keycloak usernames.

* **Tests**
  * Added coverage for available emails, matching encoded and raw usernames, and emails belonging to other users.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/IdentityManager.java
+++ b/src/main/java/de/caritas/cob/userservice/api/IdentityManager.java
@@ -69,7 +69,7 @@ public class IdentityManager implements IdentityManaging {
     var user = identityClient.findUserByEmail(email);
 
     return user.isEmpty()
-        || user.get("encodedUsername").equals(username)
-        || user.get("decodedUsername").equals(usernameTranscoder.decodeUsername(username));
+        || username.equals(user.get("encodedUsername"))
+        || usernameTranscoder.decodeUsername(username).equals(user.get("decodedUsername"));
   }
 }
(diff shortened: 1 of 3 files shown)


Storypapst:
@coderabbitai review

shazia-k (approved):

Storypapst:
Addressed the CodeRabbit stability finding in 15a9cda: equality now starts from the known non-null authenticated username values, and a new regression test verifies incomplete owner data is rejected without throwing. Focused ownership/controller tests: 21 passed.

Issue #335, opened 2026-07-08
Stabilize UserService non-blocking CI gates


Storypapst:
## Problem
PR #334 now has green required build/test checks, but the non-blocking CI gates still fail and should be made actionable.

## Current evidence
- PR: https://github.com/OpenResilienceInitiative/ORISO-UserService/pull/334
- Commit: `91a349b7ddae59b9f9b386cf20f13f31f2f68f2e`
- Required checks are green:
  - `test and build UserService`
  - `test, build and Docker validation`
- Local Java 21 validation is green:
  - `./mvnw clean test`: 2755 tests, 0 failures, 0 errors, 7 skipped
  - `./mvnw spotless:check`: success

### API contract diff
The job fails before reporting an actual API diff:

```text
failed to load base spec from /work/.base/api/appointmentservice.yaml: map key "LanguageCode" not found
```

The other specs in the same job report `No changes detected`.

### Burn-in integration tests
The burn-in job reports broad integration-test failures:

```text
Integration-test burn-in: tests=3632 failures=226 errors=169 skipped=8
```

Representative causes from the log:
- `Detached entity passed to persist: de.caritas.cob.userservice.api.model.SessionData`
- `LazyInitialization ... Consultant.consultantMobileTokens ... no session`
- `LazyInitialization ... Session.sessionTopics ... no session`
- `Connect to localhost:8080 failed: Connection refused`

## Expected fix
- Make `api-contract-diff` robust against the existing `appointmentservice.yaml` `LanguageCode` reference, or pin/fix the oasdiff input so it can load the base spec.
- Split or stabilize the burn-in integration-test profile so environmental failures (`localhost:8080`) are separated from real persistence/session failures.
- Keep the required Java 21 unit/build gate green while doing this.

Storypapst:
This issue is now the UserService implementation package under #731. It also addresses the still-open failure-propagation defect in #111 without duplicating that issue. The PR will keep existing burn-in evidence while making the stable required lane truthful.

Storypapst:
Implementation is ready for human review in #735. The new `required PreDev CI` conclusion aggregates unit/build, Redis, and deterministic integration evidence. The full historical suite remains explicitly visible under #734 until its 2026-09-30 exit criteria are met.

Storypapst:
## Delivery matrix update

| Native sub-issue | Pull request | Status |
| --- | --- | --- |
[...]

PR #744, opened 2026-07-25
test(load): exercise seeded mixed read workload

Storypapst:
## Summary

- add a deterministic weighted public-read scenario over seeded consultant, agency/topic and language data
- run two real UserService JVMs against shared isolated MariaDB 11.0.6 and Redis 7 state
- distribute traffic deterministically across both replicas and enforce aggregate, per-operation and per-replica thresholds
- exercise the generated AgencyService client through a healthy local batch-read stub
- auto-select JDK 21 on macOS and fail before Docker when no supported runtime is available
- provide bounded startup, warm-up and verified cleanup in one command

## Why

PR #735 restored truthful required CI, but the checked-in load proof exercised only liveness. The first version of this PR added seeded application reads against one UserService JVM. The current head extends that proof across two real JVMs sharing database and Redis state, without claiming that write, scheduler, authentication or deployed routing paths are already replica-safe.

The final verification also reproduced a local JDK 25 failure at Spotless/google-java-format. The runner now establishes the repository's Java 21 requirement before any container starts.

## Current two-replica evidence

`bash scripts/load/run-seeded-public-read-replicas.sh`

- 1,400 requests at concurrency 32
- 700 requests per replica
- 0 failures across all six named operations
- 405,400 response bytes
- 765.46 requests/second
- 41.54 ms mean, 80.83 ms p95, 218.58 ms maximum
- replica p95: 88.95 ms and 62.11 ms
- slowest operation p95: 102.75 ms
- no temporary containers, JVMs or listeners remained after cleanup

## Verification

- 27 Python architecture/load/CI contracts
- Python Black and compilation checks
- shell syntax for both replica scripts
- JDK 21 Maven package and Spotless
- local CodeRabbit review of both newly published commits: no findings
- previous full Java 21 unit suite: 3,795 tests, 0 failures/errors, 7 skips
- previous required integration/contract/E2E suite: 940 tests, 0 failures/errors, 3 skips

## Scope and release boundary

This is bounded local mixed-read evidence, not a production capacity claim. It does not prove concurrent writes, scheduler idempotency, authentication/authorization, Kubernetes service routing, or deployed PreDev behavior. The supported UserService replica maximum therefore remains one.

The architecture target remains Matrix-only chat, the ORISO-owned frontend, and the ORISO-controlled Element Call/MatrixRTC fork with LiveKit. Rocket.Chat and Jitsi remain removal targets; this load-test slice does not reintroduce either dependency.

Closes OpenResilienceInitiative/ORISO-UserService#743
[...]

PR #746, opened 2026-07-25
Retain anonymous deletion when error notification fails

### LINKED ISSUE #745: Prevent notification failures from rolling back anonymous deletions
## Why

Live PreDev tracing exposed a repeatable transaction-boundary defect in the anonymous-user deletion scheduler. Six runs produced 1,510 successful Matrix client calls (about 252 per run), but every root workflow failed after the secondary error-notification path tried tenant-specific template resolution in technical/global context. Because the deletion workflow is transactional, that notification exception rolled back database cleanup after irreversible Matrix side effects, causing the next run to repeat the calls.

Parent: OpenResilienceInitiative/ORISO-UserService#335

## What to build

- Make deletion error notification best-effort after deletion actions have completed.
- Preserve completed database cleanup when notification delivery or template resolution fails.
- Return the generic application URL for technical/global mail context without calling TenantService or ApplicationSettingsService.
- Log the notification failure without exposing workflow identifiers or secrets.
- Add regression tests for both the scheduler transaction boundary and the zero-outbound-call technical template path.

## Acceptance

- [x] A notification-path runtime exception does not escape the anonymous deletion scheduler.
- [x] Completed session and user cleanup remains committed.
- [x] Technical/global template rendering performs no TenantService or ApplicationSettingsService call.
- [x] The notification failure remains operationally visible in logs.
- [x] The focused regression, full Java 21 unit suite, full integration suite, formatting gate and final diff review pass.

## Runtime follow-up

Merge, PreDev deployment and trace confirmation remain part of parent stability issue OpenResilienceInitiative/ORISO-UserService#335; this code-fix issue does not claim that runtime evidence.

Implementation: OpenResilienceInitiative/ORISO-UserService#746

Storypapst:
## Summary

- keep completed anonymous-user cleanup committed when secondary error notification fails
- skip tenant-specific template lookups in technical/global context, eliminating the observed TenantService 404 path
- log only the notification failure type, without workflow identifiers or dependency exception details
- add transaction-boundary and no-outbound-call regressions
- record the live PreDev trace baseline and build/deploy/runtime separation

## Why

Read-only PreDev traces showed six anonymous-deletion scheduler runs making 1,510 successful Matrix client calls before every root workflow failed in tenant-template resolution. The notification exception escaped the outer transaction, rolling back database cleanup after irreversible Matrix side effects and causing the next scheduler run to repeat those calls.

The trace also identified the upstream dependency error: technical/global mail rendering attempted tenant-specific branding through TenantService even though that context has no tenant. The canonical tenant-template supplier now returns only the generic application URL in that context and does not call TenantService or ApplicationSettingsService.

## Delivery order

This is the small, review-ready immediate repair for the exact observed RuntimeException and technical-tenant lookup path.

Draft PR #759 is the follow-up structural hardening: it gives deletion its own transaction boundary, preserves notification failure visibility, and makes Keycloak `Unauthorized → NotFound` idempotent. #759 must be rebased after both this PR and #757 land; it is not an independent alternative to merge in parallel.

## Verification

- Red transaction regression: failed with `IllegalStateException: error notification failed`
- Red supplier regression: technical context called TenantService and failed while resolving tenant data
- Focused supplier test: 4 tests, 0 failures, 0 errors
- Full Java 21 unit suite: 3,801 tests, 0 failures, 0 errors, 7 skipped
- Full integration/contract/E2E suite: 941 tests, 0 failures, 0 errors, 3 skipped
- Spotless: clean
- `git diff --check`: clean
- CodeRabbit CLI review of the final uncommitted diff: 0 findings
- CI on `80ce2d5f`: branch and PR Java/build, Docker, required integration, Redis, both MariaDB contracts, API diff and aggregate PreDev gate all green
- GitHub review threads: 0 unresolved; human review still pending

## Runtime boundary

This PR has local and trace-backed evidence only. It does not claim merge, deployment, running-image, custom-metric, or post-deploy scheduler proof.

## Scope correction

This no longer claims to close #745. The regression test now provokes the
genuine database-origin failure instead of stubbing it, and that reproduction
shows the repair is partial: the notification exception is caught as intended,
but Hibernate has already marked the transaction rollback-only, so the commit
still fails with `UnexpectedRollbackException` and no deletion is retained.

What this PR does remove is the most frequent trigger — the tenant-template
lookup in the technical/global context — and the notification exception
escaping the transaction. The remaining gap needs the deletion to own its
transaction boundary, which is #759. Expect sessions `103302`/`103430` to
survive the deploy of this branch.

Refs OpenResilienceInitiative/ORISO-UserService#745
Refs OpenResilienceInitiative/ORISO-UserService#758
Refs OpenResilienceInitiative/ORISO-UserService#759
Refs OpenResilienceInitiative/ORISO-UserService#335


kiodreambau (commented):
Reviewed against live PreDev evidence collected 2026-07-27. **This should be merged** — it removes a defect that is provably firing every hour — but it will not on its own satisfy the runtime acceptance criterion in #745/#335, and the regression test does not cover the failure mode that is actually happening.

## The change is correct and the defect is real

Confirmed on PreDev, one trace, pod `oriso-platform-userservice-665446b65f-jk5cq`, run of 2026-07-27T22:00:

```
HttpClientErrorException$NotFound: 404 on GET .../tenant/public/id/0
  at TenantTemplateSupplier.getTemplateAttributes(TenantTemplateSupplier.java:41)
  at WorkflowErrorMailService.buildAndSendErrorMail(WorkflowErrorMailService.java:50)
  at DeleteUserAnonymousService.deleteInactiveAnonymousUsers(DeleteUserAnonymousService.java:41)
  at DeleteUserAnonymousService$$SpringCGLIB$$0.deleteInactiveAnonymousUsers(<generated>)
  at DeleteUserAnonymousScheduler.performDeletionWorkflow(DeleteUserAnonymousScheduler.java:22)
```

The exception escapes through the `@Transactional` proxy, so the whole run rolls back. Database read-back proves the consequence: sessions `103302` (last touched 2026-07-18) and `103430` (2026-07-20) are still present, after roughly 200 hourly runs that each "deleted" them. The scheduler is hourly (`user.anonymous.deleteworkflow.cron=0 0 * * * ?`), not nightly, so this has been looping about 24x per day.

Both halves of the fix are right: the `try/catch` around the notification, and short-circuiting `TenantTemplateSupplier` for the technical/global context so the invalid tenant-0 lookup is not made at all.

## Gap 1 — the regression test does not reproduce the live failure

`performDeletionWorkflow_Should_commitDeletion_When_errorNotificationDependencyFails` produces the workflow error by throwing from `MatrixSynapseService.deactivateUser`, i.e. an **external** call. On PreDev the workflow error originates **inside the database delete**:

```
ObjectNotFoundException: No row with the given identifier exists for entity
  [de.caritas.cob.userservice.api.model.Session with id '103430']
  at DefaultMergeEventListener.entityIsDetached(DefaultMergeEventListener.java:458)
  at DeleteDatabaseAskerAction.execute(DeleteDatabaseAskerAction.java:36)
```

`userRepository.delete(user)` merges the detached `User`; Hibernate's `TypeHelper.replace` then walks the already-initialized `User.sessions` collection and tries to re-resolve the sessions that `DeleteAskerRoomsAndSessionsAction` deleted moments earlier in the same run.

That matters for two reasons:

1. In the test the user delete **succeeds**, so `assertSessionAndUserDoNotExistInDatabase` passes trivially. On PreDev the user delete is precisely what fails, and it will keep failing after this merge — sessions will start committing, user rows will not.
2. A DB-origin exception can leave the persistence context marked rollback-only, in which case catching the notification exception does not save the commit. An HTTP 404 cannot do that, so the test never exercises this.

Please add a second case where the **database** delete throws, asserting that whatever completed still commits. That is the case #745 actually needs to be true.

## Gap 2 — scope note, not a blocker

The transaction still spans the entire batch plus the irreversible Matrix and Keycloak calls. This PR removes one trigger; it does not make a single bad user harmless to the other users in the same run. #824 is the right shape for that (see my review there).

## Verdict

Merge, deploy, then verify at the next `:00` run against sessions `103302`/`103430`. If they disappear, the rollback loop is closed and this PR is confirmed. Expect the `group-chat-system-*` **user** rows to survive — that is Gap 1 and needs its own issue.

kiodreambau:
## Follow-up: the database-origin case is now verified locally

In my review above I flagged that the regression test only exercises an **external** failure source, and that a DB-origin workflow error might still mark the persistence context rollback-only. I ran that case on this branch rather than leaving it open.

Probe added to `DeleteUserAnonymousSchedulerIT` on `pr-746`: `UserRepository.delete` is stubbed to throw `ObjectRetrievalFailureException` (the exact type PreDev produces), the mail service throws as well, then the database is read back.

```
PROBE session present after run = false
PROBE user present after run    = true
```

`Tests run: 6, Failures: 0, Errors: 0` — Java 21, offline Maven, ~20 s per run.

**Two conclusions:**

1. **This PR does fix the real failure mode.** `ObjectRetrievalFailureException` does not mark the transaction rollback-only, so the completed session deletion commits. The hourly rollback loop on PreDev closes with this merge. That resolves the doubt in my review — no additional change is needed for #745's transaction-boundary criterion.

2. **The residue is exactly as predicted.** The user row survives. So the falsifiable prediction for PreDev, checkable at the first `:00` run after deployment:

   | Row | Expected after deploy |
   | --- | --- |
   | session 103302, 103430 | **gone** |
   | user `group-chat-system-1`, `group-chat-system-84` | **still present** |

   If both hold, this PR is confirmed and the remaining defect is isolated to `DeleteDatabaseAskerAction` (merge over the stale `User.sessions` collection). If the sessions also survive, my analysis is wrong and I will re-open it.

The probe is scratch and deliberately not pushed — it prints rather than asserts, so it documents behaviour instead of pinning it. Say the word and I will turn it into a proper assertion-based regression test on this branch.

kiodreambau:
## Regression test added, branch brought onto current `pre-dev` — now green

Pushed `9fc4ddd8` (regression test) and `f005b929` (merge of `pre-dev`, no conflicts). All checks pass, `MERGEABLE CLEAN`.

### The regression test

`performDeletionWorkflow_Should_commitSessionDeletion_When_userDeleteFailsInDatabase` stubs `UserRepository.delete` to throw `ObjectRetrievalFailureException` — the exact type PreDev produces — fails the notification on top, and asserts the completed session deletion stays committed while the user row is left intact rather than silently lost.

Red/green verified locally before pushing:

| | Result |
| --- | --- |
| Without the fix (main sources reverted to `pre-dev`) | **FAIL** — `Unexpected exception thrown: IllegalStateException: error notification failed` |
| With the fix | **PASS** — 6 tests, 0 failures, 0 errors |

This closes the gap I raised in my review: the existing test only drove the error path from an external call, so it could not show whether a database-origin exception leaves the persistence context rollback-only. It does not.

### Heads-up: this branch was not actually green any more

`openapi provider and consumer contracts` failed on my first push. That job **did not exist** on the previous head `80ce2d5f` (25 Jul) — it arrives through the `pull_request` merge ref from `pre-dev`, and it flagged the Rocket.Chat teardown removals (`rctoken`, `rcuserid`, `rcGroupId`) because this branch's provider spec was 78 commits behind. Merging `pre-dev` resolved it.

Worth knowing generally: **any PR in this lane whose last green run predates that gate is not as green as `gh pr checks` suggests** until it is rebased.

### Separate finding: `Closes` does not link in the `pre-dev` lane

The body carries `Closes OpenResilienceInitiative/ORISO-UserService#745`, but `closingIssuesReferences` is empty — and it is empty for #824, #744, #755 and #753 as well. GitHub only populates that relation for PRs targeting the default branch (`main`), so nothing in the `pre-dev` lane auto-links or auto-closes. Issue status has to be moved by hand on merge. Raising it here because the ORISO issue workflow assumes the opposite.

### Follow-up filed

The residue this PR deliberately does not fix — the user row surviving — is now #859 with sub-issue #860.

Issue #748, opened 2026-07-25
Rename the rc* wire fields to their Matrix meaning (cross-repo)


Storypapst:
## Why

Four field names survive from Rocket.Chat and are still **live API contract**, so they cannot be renamed in one repo alone:

| Field | What it actually holds today | Where |
|---|---|---|
| `rcGroupId` | the **Matrix room id** (`CreateEnquiryMessageFacade:220` writes `.rcGroupId(matrixRoomId)`) | REST DTOs, STOMP video-call payload, `SecurityConfig` route matcher, DB column `Session.groupId` |
| `rcUserId` | the ban/chat user id | `banUser` endpoint path |
| `rcToken` / `RCToken` header | still required by the archive endpoints | `fetchData.ts` sends it when `rcValidation` is set |
| `rcValidation` | frontend flag that decides whether to attach the RC headers | `fetchData.ts` |

The frontend purge (ORISO-Frontend#784) removed every *internal* Rocket.Chat reference and deliberately stopped at this boundary — renaming these there would have broken the API.

## Proposed target names

- `rcGroupId` → `groupId` (matches the entity field `Session.groupId` already) or `matrixRoomId` if we want the meaning explicit
- `rcUserId` → `chatUserId`
- `RCToken` / `RCUserId` headers → drop entirely if the archive endpoints no longer validate them; otherwise rename

**First step is a question, not a rename:** do the archive endpoints still actually validate `RCToken` / `RCUserId`, now that Rocket.Chat is off? If not, the header pair and `rcValidation` can be deleted rather than renamed, which removes 69 occurrences from the frontend on its own.

## Sequencing

1. Answer the `RCToken` question above; delete if dead.
2. UserService: add the new field name alongside the old one in the OpenAPI spec, serve both.
3. Frontend + Admin: regenerate types, switch to the new name.
4. UserService: drop the old name.

Related: ORISO-UserService#747 (delete the adapter), ORISO-Frontend#784 (frontend purge).

### Reviewer test plan

- [ ] Compare UserService and Frontend generated contracts for matrixRoomId and matrixUserId with no uncoordinated rc field dependency.
- [ ] Exercise serialization and migration compatibility against representative existing rows.
- [ ] Deploy UserService and Frontend in one Helm release and reject any mixed-version DTO failure.

Storypapst:
## The opening question is answered: `RCToken` is dead, `RCUserId` is not

[...]

Issue #747, opened 2026-07-25
Complete the Matrix-only UserService cutover and delete Rocket.Chat


Storypapst:
Parent: OpenResilienceInitiative/ORISO-Frontend#302
Related contract cleanup: OpenResilienceInitiative/ORISO-UserService#748

## Why
Rocket.Chat is disabled at runtime but its adapter, credentials, identifiers, API contracts, schema columns, fixtures, and lifecycle branches still exist. The complete Matrix-only replacement is already implemented locally and must be reviewed atomically with the frontend and runtime cutover.

## What to build
Publish the complete UserService cutover: make enquiry, account, consultant, session-list, group-chat, assignment, archive, notification, and deletion flows Matrix-native; delete the Rocket.Chat adapter and configuration; remove obsolete rc* API/schema contracts; keep Matrix room ownership and access checks explicit; and publish an immutable image.

## Acceptance
- [x] Active source, tests, API contracts, and runtime configuration contain no Rocket.Chat adapter or credentials.
- [x] Session and group-chat lifecycle operations use Matrix-only services and neutral room identifiers.
- [ ] The full Java 21 unit and integration suites pass without Rocket.Chat fixtures.
- [ ] The coordinated frontend contract consumes the Matrix-only API.
- [ ] The merged pre-dev image is pinned by digest and verified in registration, counselling, group-chat, and deletion journeys.

### Reviewer test plan

- [ ] Run the Java 21 package and Matrix lifecycle contract gates, including session, identity, and admin paths.
- [ ] Inspect source, generated contracts, migrations, and runtime configuration for reachable Rocket.Chat or LiveService transports.
- [ ] Verify the reviewed immutable image and database state on PreDev before marking the issue QA-complete.

Storypapst:
## This is already done on `pre-dev` — measured, not assumed

I went to do this and found the work had landed. Evidence:

| check | result |
|---|---|
| `rocketchat` in `src/main/**/*.java` | **0 files** |
| `api/adapters/rocketchat`, `api/exception/rocketchat` | gone |
| `rcGroupId` / `rcUserId` / `rcToken` in **any** `api/*.yaml` | **0** — replaced by `matrixRoomId` (17 occurrences) |
| DB columns | changeset `0074_remove_rocket_chat_room_ids`, with a `legacy_chat_identifier_archive` table and `onFail="HALT"` |

That last one also covers #748, which can be closed with this.

## What is left, and why it should stay

[...]

PR #751, opened 2026-07-26
Make the RCToken header optional everywhere

Storypapst:
## The finding

`RCToken` is a Rocket.Chat leftover that carries no meaning any more — but four endpoints still returned **400** without it.

Traced end to end:

1. `UserSessionControllerDelegate` builds `RocketChatCredentials` from the header and, when it is absent, substitutes the literal **`"dummy-rc-token"`** (line 295). A token that may be replaced by a placeholder validates nothing.
2. Those credentials reach `RocketChatRoomInformationProvider.retrieveRocketChatInformation`, which opens with `if (!rocketChatEnabled || isMatrixMigrationCredential(...))`.
3. `rocket-chat.enabled` is `false` by default (ADR-004), so that branch **always** wins — and it reads only `getRocketChatUserId()`. **The token is never read.**

Meanwhile the frontend sends `MATRIX_MIGRATION_DUMMY_RC_TOKEN`. Both sides were feeding each other a placeholder to satisfy a check that no longer exists.

## The change

`required: true` → `required: false` for `RCToken` in `api/userservice.yaml` (3) and `api/conversationservice.yaml` (2), plus `required = false` on the four `@RequestHeader` declarations.

Relaxing the flag moves the parameter **behind** the required ones in the generated interfaces, so three controller signatures had to be reordered to keep overriding — `getChatById`, `getSessionsForAuthenticatedConsultant`, `getTeamSessionsForAuthenticatedConsultant`.

This is a pure relaxation: a client that still sends the header is unaffected.

## Not touched: `RCUserId`

That one is **still live**. It becomes `RocketChatCredentials.rocketChatUserId` and is read for the Matrix room lookup. Only the token is dead. Renaming it belongs to #748.

## Testing

| Suite | Result |
|---|---|
| `UserControllerIT` | 135/135 |
| `ConversationController*` + `*SessionList*` | 104/104 |

Two IT cases asserted the old contract (`..._Should_ReturnBadRequest_WhenHeaderParamIsMissing`) and now assert the new one, renamed accordingly.

## Build note worth recording

The test suite **only runs on JDK 21**. On JDK 25 all 557 selected tests error out with `MockitoException: Mockito cannot mock this class` before a single assertion runs. I nearly misread that as a regression — it is the local toolchain. Worth pinning in CI docs.

Related: #747, #748.

### DIFF
--- a/api/conversationservice.yaml
+++ b/api/conversationservice.yaml
@@ -116,7 +116,7 @@ paths:
             type: integer
         - name: RCToken
           in: header
-          required: true
+          required: false
           schema:
             type: string
       responses:
@@ -160,7 +160,7 @@ paths:
             type: integer
         - name: RCToken
           in: header
-          required: true
+          required: false
           schema:
             type: string
       responses:
--- a/api/userservice.yaml
+++ b/api/userservice.yaml
@@ -747,7 +747,7 @@ paths:
       parameters:
         - name: RCToken
           in: header
-          required: true
+          required: false
           schema:
             type: string
         - name: status
@@ -855,7 +855,7 @@ paths:
       parameters:
         - name: RCToken
           in: header
-          required: true
+          required: false
           schema:
             type: string
         - name: offset
@@ -1661,7 +1661,7 @@ paths:
       parameters:
         - name: RCToken
           in: header
-          required: true
+          required: false
           schema:
             type: string
         - name: chatId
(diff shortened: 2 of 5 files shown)


Storypapst:
@coderabbitai review

Ready for review. The core claim to check is the dead-code argument for `RCToken`: `UserSessionControllerDelegate:295` substitutes the literal `"dummy-rc-token"` when the header is absent, and `RocketChatRoomInformationProvider.retrieveRocketChatInformation` opens with `if (!rocketChatEnabled || isMatrixMigrationCredential(...))` — with `rocket-chat.enabled=false` that branch always wins and reads only `getRocketChatUserId()`. If that reasoning is wrong anywhere, the relaxation should not land.

Note the test suite only runs on JDK 21; on JDK 25 every test errors with `MockitoException: Mockito cannot mock this class` before a single assertion.

shanzaeimran (approved):

PR #757, opened 2026-07-26
feat(stability): make UserService replica safety measurable (#543)

Storypapst:
## Why

UserService runs as a Kubernetes Deployment but still combines replica-local state, scheduled side effects and external provider coordination. A Deployment alone is not evidence that requests and jobs are safe across multiple replicas, and the missing proof made operational risk difficult to measure.

## What changed

- inventories replica-local state and publishes bounded replica-safety, scheduler and outbound dependency metrics;
- adds finite timeout and payload/latency instrumentation for generated HTTP clients and the Keycloak admin client;
- externalizes correctness-relevant caches, one-time tokens, notification active-view state, Matrix browser login coordination and Matrix sync leadership/cursors through Redis;
- adds durable scheduler claims, database idempotency constraints and two-instance MariaDB/Redis contracts;
- makes the required CI gates execute the complete integration suite plus dedicated MariaDB and Redis contracts;
- neutralizes identity creation, login and profile lookup ports so application code no longer consumes Keycloak response or SDK types;
- records reproducible unit, integration, local image, load and degradation evidence.

## Developer and operational impact

The runtime ceiling intentionally remains one supported replica. This PR makes the remaining blockers explicit and measurable; it does not authorize a Helm replica increase. Inactive-account recovery also remains disabled until the deployed MailService proves idempotent replay with the same key.

## Dependencies and architecture boundary

- **Merge gate:** #740 must land in `pre-dev` first. This branch already integrates its functional commits and must not be merged or deployed while #740 remains open.
- **Matrix-only cutover:** the target architecture is full Rocket.Chat removal with Matrix, the ORISO-owned frontend, the controlled Element Call/MatrixRTC fork and LiveKit, plus removal of legacy Jitsi dependencies. That physical removal is tracked in the coordinated Matrix-only package, including UserService #756; this stability PR does not legitimize a disabled Rocket.Chat fallback as an end state.
- **Remaining proof:** authenticated PreDev/SigNoz readback, deployed MailService idempotency and complete two-instance product E2E without sticky sessions are still required before the replica ceiling can change.

## Validation

- `./mvnw -B -Dskip.integration-tests=true clean test`: 3,843 tests, 0 failures, 0 errors, 7 skipped
- `ORISO_LOCAL_REDIS_IT=true ./mvnw -B -Dskip.unit-tests=true clean integration-test`: 964 tests, 0 failures, 0 errors, 5 skipped
- required fresh MariaDB 10.11 contract: 9 tests, 0 failures
[...]

PR #774, opened 2026-07-27
fix(replica): make authenticated tutorial writes atomic (#773)

### LINKED ISSUE #773: Prove authenticated writes across two UserService replicas
## Why

The existing replica work proved bounded public reads and individual Redis/MariaDB coordination modules, but not a valid JWT-authenticated write routed between replicas. Tutorial progress also implemented upsert as read-then-insert, so concurrent replicas could race on the unique scope key. Concurrent creation of different scopes could separately exceed the per-user row cap.

Parent: #733

## Delivered in #830

- deterministic local JWK/JWT adapter for the real resource-server authentication path
- two packaged UserService JVMs against shared isolated MariaDB and Redis
- native MariaDB atomic upsert for one tutorial-progress scope
- hashed, user-scoped advisory serialization for first writes so the row cap remains atomic across replicas
- existing-scope lock-free update path
- portable H2 HTTP/JPA E2E plus separate real-MariaDB concurrency contracts
- alternating authenticated PUT/GET traffic across both replicas plus one-replica restart

## Acceptance

- [x] A real MariaDB concurrency test covers concurrent first writes and leaves one canonical scope.
- [x] A real MariaDB concurrency test proves different new scopes cannot exceed the per-user row cap.
- [x] The rejected-write path releases the advisory lock for the next first write.
- [x] Two real JVMs accept a valid consultant JWT without sticky sessions or a real Keycloak dependency.
- [x] Concurrent same-scope writes return no 5xx and leave exactly one canonical progress row.
- [x] Writes are immediately readable from the other replica.
- [x] Stopping and restarting one replica does not lose authentication or persisted progress.
- [x] The report enforces per-operation and per-replica volume, error-rate and p95 thresholds.
- [x] The runner removes every temporary JVM, listener and container.
- [x] Exact-head GitHub Unit, Required Integration, MariaDB, Redis and OpenAPI gates are terminal green.

## Current evidence

PR: #830, exact local head `73c60742`

- focused tutorial unit suite: 9 tests, 0 failures
- tutorial controller E2E: 5 tests, 0 failures
- complete required integration suite: 79 reports / 844 tests, 0 failures, 0 errors, 7 environment-gated skips
- real tutorial MariaDB contract: 4 tests, 0 failures
- initial two-JVM phase: 82 authenticated operations, 0 failures, 41.76 ms p95
- restart phase after isolated-scope initialization: 14 measured operations, 0 failures, 19.22 ms p95
- one canonical measured-scope MariaDB row and no duplicate-key warning
- per-operation and per-replica 1,000 ms p95 threshold remained enforced

This is bounded local branch evidence for the tutorial-progress state transition. It is not a claim that #830 is merged, deployed to PreDev, proven through deployed SigNoz, or sufficient to raise the whole-service replica limit.

Storypapst:
## Summary

- move tutorial-progress persistence behind a dedicated transactional store
- replace read-then-insert with MariaDB `ON DUPLICATE KEY UPDATE`
- use `READ_COMMITTED` for the bounded upsert transaction so identical concurrent no-op updates can immediately read the canonical winner
- add a real MariaDB race contract that proves two instances return one row without duplicate-key warnings
- add a reusable two-JVM authenticated load/restart proof with a disposable JWK/JWT endpoint, shared MariaDB and Redis
- run that proof from the required MariaDB workflow and document the bounded slice

## Root cause

Two replicas could both read a missing tutorial-progress scope and then race to insert the same unique key. Catching the duplicate avoided one 500 but still emitted warning stacks. A native upsert removed that exception path, but real HTTP load exposed a second detail: MariaDB stores these timestamps at one-second precision, so an identical `ON DUPLICATE KEY UPDATE` can be a no-op. Under the default `REPEATABLE_READ` snapshot, the transaction could then fail to see the row committed by the winning replica. The small store-owned transaction now uses `READ_COMMITTED` and reads the canonical row after the atomic statement.

## Runtime proof

The local proof packaged the real application and started two UserService JVMs against one disposable MariaDB 11.0.6 database and shared Redis 7. A locally signed consultant JWT was verified through the real resource-server JWK path; every PUT also passed the real CSRF filter. A separate version-999 scope warms the fresh JWT/security/controller/SQL paths and is excluded from the report. The measured version-1 scope remains absent until the main phase, so its first-write race is still exercised.

- initial phase: 80 concurrent PUTs alternating across both replicas plus one GET per replica; 82 requests, 0 failures, 39 ms aggregate p95
- restart phase: restart replica one, then 12 further PUTs plus one GET per replica; 14 requests, 0 failures, 372 ms aggregate p95
- per-operation and per-replica zero-error and 1,000 ms p95 gates passed
- exactly one canonical measured-scope database row remained
- both replicas read the state and no duplicate-key warning was emitted
- all temporary JVMs, listeners and containers were removed

## Validation

- Java 21 unit suite: 3,812 tests, 0 failures, 0 errors, 7 skipped
- required integration/E2E suite with Redis contracts: 967 tests, 0 failures, 0 errors, 8 skipped
- real MariaDB schema/replica selection: 12 tests, 0 failures, 0 errors, 0 skipped
- Python CI contracts: 25 tests passed
- shell/Python syntax, Spotless and `git diff --check` passed
- GitHub Feature Branch and Pull Request workflows: all terminal checks passed on `85eeff1d`

## Dependencies and scope

- stacked on #757 (`feat/replica-safety-metrics-543`)
- complements the public-read load evidence in #744
- proves only the authenticated tutorial-progress state transition; it does not raise the whole-service replica limit and is not deployed-runtime or SigNoz evidence
- preserves the Matrix-only target architecture: no Rocket.Chat or Jitsi fallback is introduced; the proof explicitly disables Rocket.Chat and the Matrix listener

Closes #773

### DIFF
--- a/.github/workflows/mariadb-contract.yml
+++ b/.github/workflows/mariadb-contract.yml
@@ -38,5 +38,11 @@ jobs:
     - name: Verify real MariaDB schema contracts
       run: >-
         ./mvnw -B -Dskip.unit-tests=true
-        -Dtest=DatabaseChangelogDriftIT,AdminStatisticsRepositoryMariaDbIT,OrganizerMariaDbReplicaIT,DeactivateGroupChatSchedulerMariaDbReplicaIT,ConsultantMessageStatServiceMariaDbReplicaIT,InactiveAccountNotificationServiceReplicaIT,DeleteInactiveSessionsAndUserSchedulerReplicaIT
+        -Dtest=DatabaseChangelogDriftIT,AdminStatisticsRepositoryMariaDbIT,OrganizerMariaDbReplicaIT,DeactivateGroupChatSchedulerMariaDbReplicaIT,ConsultantMessageStatServiceMariaDbReplicaIT,InactiveAccountNotificationServiceReplicaIT,DeleteInactiveSessionsAndUserSchedulerReplicaIT,TutorialProgressServiceMariaDbReplicaIT
         integration-test
+
+    - name: Prove authenticated writes across two replicas
+      env:
+        USERSERVICE_LOAD_REQUESTS: 80
+        USERSERVICE_LOAD_CONCURRENCY: 12
+      run: ./scripts/load/run-authenticated-write-replicas.sh
--- a/documentation/USER_SERVICE_REPLICA_SAFETY.md
+++ b/documentation/USER_SERVICE_REPLICA_SAFETY.md
@@ -236,6 +236,29 @@ database reads, provider cleanup or error-mail work.
 one context setup and one invocation of each mode under two concurrent
 scheduler instances. Its 12-hour claim has the same daily duration constraint.
 
+Authenticated tutorial-progress writes no longer implement read-then-insert in
+the service layer. `TutorialProgressStore` owns the transaction and uses
+MariaDB's native `ON DUPLICATE KEY UPDATE` against the unique
+user/surface/tour/version scope. Its small transaction uses `READ_COMMITTED`:
+when two replicas send an identical update within the same `DATETIME` second,
+the database may perform a no-op update, and a `REPEATABLE_READ` snapshot would
+otherwise remain unable to see the winning row. The real-MariaDB
+`TutorialProgressServiceMariaDbReplicaIT` forces both instances to read the
+missing scope before either writes and proves both calls return the one
+canonical row without a duplicate-key warning. It also preserves the per-user
+row cap for new scopes while allowing existing scopes to update.
+
+`scripts/load/run-authenticated-write-replicas.sh` extends that database proof
+through the running HTTP/security stack. It starts two real UserService JVMs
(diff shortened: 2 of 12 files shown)


Storypapst:
Superseded by #830, replayed directly on the current Matrix-only `pre-dev` baseline. The replay keeps only the tutorial atomic-write implementation, MariaDB race contract, and authenticated two-replica load/restart proof; it does not restore the old stacked replica documentation or unrelated contract list. The new runner also removes the legacy `ROCKET_CHAT_ENABLED` setting entirely.

kiodreambau:
Superseded by #830, which replays this slice onto current `pre-dev`. That PR declares the supersession explicitly in its own body. Closing here so the open-PR queue reflects the replay lane rather than the abandoned stack; the branch is untouched and this PR can be reopened if the replay is dropped. Refs #335

Issue #758, opened 2026-07-26
Prevent deletion notification failures from replaying external side effects


Storypapst:
Parent: #543

## Runtime evidence

On PreDev, the anonymous-user deletion scheduler repeatedly processes the same batch. A measured execution performed 141 Matrix user deactivations and 70 room purges. The admin token was reused correctly: only one Matrix admin login occurred.

The trigger is a Keycloak 404 after an unauthorized-session refresh. That workflow error causes error-mail rendering to query technical tenant 0, which returns 404. Because notification currently runs inside the deletion transaction, the database work rolls back after the external Keycloak and Matrix side effects have already completed.

## Scope

- Treat Keycloak NotFound after the single unauthorized-session retry as idempotent success.
- Commit the anonymous-user deletion transaction before workflow-error notification.
- Preserve workflow error reporting without allowing mail or tenant lookup failures to roll back completed deletion work.
- Add focused regression tests for both failure paths.
- Record the measured before/after fan-out contract.

## Acceptance criteria

- [x] Unauthorized then NotFound during Keycloak deletion is swallowed as already deleted.
- [x] A workflow-error notification failure cannot roll back the deletion transaction.
- [x] Focused unit and integration tests pass.
- [x] Full UserService unit and integration suites remain green.
- [ ] After deployment, the same users are not deactivated or purged again in the next scheduler execution.

## Delivery order

- #746 is the immediate fix for the observed notification RuntimeException and invalid technical-tenant lookup.
- #757 provides the replica-safety and observability baseline on which draft #759 is currently stacked.
- #759 is the structural follow-up and must be rebased after #746 and #757 land.

Required sequence: #746 → #757 → rebase/retarget #759 → human review → merge → PreDev rollout → SigNoz replay check.

Storypapst:
Fresh read-only PreDev SigNoz baseline (last 24h, 2026-07-27): two anonymous-deletion scheduler traces each repeated exactly 141 Matrix user deactivations and 70 room purges. Across both runs that is 282 deactivation calls but only 141 unique targets, and 140 purge calls but only 70 unique rooms, confirming full replay while #759 remains undeployed. The running UserService pod is 16h old. This is baseline evidence only; the acceptance criterion still requires a post-deploy next-run comparison proving the targets are not selected again.

Storypapst:
[...]

PR #759, opened 2026-07-26
fix(deletion): prevent external side-effect replay

Storypapst:
## Summary

- commit the anonymous-user deletion batch before workflow-error notification starts
- treat Keycloak `NotFound` after the single unauthorized-session refresh as an idempotent already-deleted outcome
- add unit and Spring/H2 regression coverage for notification failure after external cleanup
- document the live SigNoz evidence and the transaction-ordering guarantee

## Runtime evidence

Read-only PreDev traces showed the same 141 Matrix user deactivations in every measured anonymous-deletion run. The failure chain was:

1. Keycloak returned unauthorized and then `NotFound` after the one allowed refresh.
2. Matrix cleanup and database deletion continued.
3. Workflow-error notification failed during a technical-tenant lookup.
4. Because notification still ran inside the deletion transaction, database state rolled back after the external side effects and the same users were selected again.

## Relationship to #746

PR #746 is the smaller immediate repair for the exact observed RuntimeException and technical/global tenant-template lookup. This draft is the structural follow-up:

- #746 catches the observed notification RuntimeException and removes the invalid tenant lookup.
- #759 ends the deletion transaction before notification begins, so later notification failures cannot roll back deletion even if their failure shape changes.
- #759 retains notification failure visibility instead of treating the scheduler execution as fully successful.
- #759 additionally makes Keycloak `Unauthorized → NotFound` an idempotent already-deleted outcome.

These PRs must not be merged in parallel. Merge/review #746 first, then #757, then rebase and retarget this draft to the resulting `pre-dev`.

## Regression proof

The Spring/H2 regression test forces a Matrix workflow error and a subsequent notification exception, then proves the user and session deletion remain committed.

## Architecture boundary

This repair supports the complete migration target: remove Rocket.Chat and Jitsi, use Matrix as the sole chat transport, retain the ORISO-owned frontend, and use an ORISO-controlled Element Call/MatrixRTC fork with LiveKit for calls. It does not claim that the coordinated migration package is already merged or deployed.

## Verification

- 3,849 unit tests: 0 failures, 0 errors, 7 skipped
- 965 integration/contract/E2E tests: 0 failures, 0 errors, 5 skipped
- 25 Python architecture/CI contracts passed
- Spotless check passed across 1,276 Java files
- `git diff --check` passed
- all branch, PR, Redis, MariaDB, API and aggregate PreDev GitHub gates passed

## Dependencies

- Depends on #746 for the immediate technical-context repair
[...]

PR #824, opened 2026-07-27
fix(deletion): replay external side-effect idempotency (#758)

Storypapst:
## Summary

Isolate each anonymous-user deletion in its own transaction, so one user can no longer end the whole run, and filter the workflow to actually anonymous users.

## The defect

The deletion batch ran in one transaction. A genuine database error marks its persistence context rollback-only, and every other user's completed work was then discarded — while their Matrix and Keycloak deletions had already happened. The next hourly run repeated those irreversible external calls.

#746 moved notification out of the transaction and removed the tenant-404 trigger. That closed the notification path, not the rollback. Its regression test pins exactly that: a database-origin failure still lost the whole batch.

## The change

Selection, deletion and notification each own their scope:

| Class | Transaction | Responsibility |
| --- | --- | --- |
| `AnonymousUserDeletionCandidates` | read-only | selects candidates, returns user **ids** |
| `AnonymousUserDeletionUnit` | `REQUIRES_NEW` | deletes one user, commits on return |
| `AnonymousUserDeletionBatch` | none | catches a failed per-user commit, records it, continues |
| `DeleteUserAnonymousService` | none | notifies after the batch |

Ids rather than entities is deliberate. An entity handed across a transaction boundary is detached in the next one, and that is precisely what made the delete take Hibernate's merge path over an already-initialized session collection — the `ObjectNotFoundException` recorded in `documentation/USER_SERVICE_STABILITY.md`. Loading the user inside the deleting transaction removes that path.

A poisoned user still cannot be retained; its own transaction is doomed by definition. It is now the only one lost, and it is reported as a workflow error instead of failing silently.

## Review points from the draft review

**Blocking-ish 1 — batch-wide transaction.** Addressed as above: the boundary is now per user.

**Blocking-ish 2 — no registration filter.** Addressed: selection uses `findByStatusInAndRegistrationType(Set.of(DONE), ANONYMOUS)`. Without it this workflow, configured under `user.anonymous.deleteworkflow.*`, also selected registered system accounts such as the per-tenant `group-chat-system-*` users — the two rows the hourly job had been failing to delete since 18 July.

**`user_chat` RESTRICT FK with no delete step.** No longer open. `DeleteDatabaseAskerAction` deletes chat memberships and mobile tokens before the user (commit `5ad3380e`, already on `pre-dev`). The argument for per-user isolation stands regardless — any `DataIntegrityViolationException` has the same effect.

## Regression proof

`DeleteUserAnonymousSchedulerIT` provokes a real `DataIntegrityViolationException`: one user keeps a `user_agency` row, so that user's delete violates the restricting foreign key at flush and its commit fails for real. The test then asserts the other user in the same batch is still deleted and committed, and that the failure is reported.

Stubbing an exception at the repository boundary would reproduce the shape of that failure but leave the persistence context healthy — such a test passes and certifies a fix that does not hold. The failure is therefore provoked, never injected, and a guard assertion fails if the workflow error stops originating in the database delete.

## Verification

- 3,439 unit tests: 0 failures, 0 errors, 0 skipped
- 853 required integration tests across 83 reports: 0 failures, 0 errors, 9 skipped
- Spotless check clean, `git diff --check` clean
- Rebased onto `pre-dev` after #746 merged

## Runtime boundary

Local and code evidence only. Deployment and post-deploy scheduler proof remain open. After rollout, sessions `103302`/`103430` disappearing is the signal this actually closed the loop.

Closes OpenResilienceInitiative/ORISO-UserService#758
Refs OpenResilienceInitiative/ORISO-UserService#745
Refs OpenResilienceInitiative/ORISO-UserService#746


kiodreambau (commented):
Reviewed together with #746 against live PreDev evidence from 2026-07-27. The direction here is better than #746 — extracting the transaction into `AnonymousUserDeletionBatch` so notification runs after commit is the structurally right move, and the Keycloak change is a genuine bug fix. Two things should change before this leaves draft.

## What is right

`AnonymousUserDeletionBatch.deleteOverdueUsers()` is a separate bean, so the `@Transactional` proxy actually applies when `DeleteUserAnonymousService` calls it. The transaction ends before `buildAndSendErrorMail`, which is the correct boundary.

`removeUserIfPresent` fixes a real defect: in the old code the retry after `NotAuthorizedException` called `remove()` again **outside** any `NotFoundException` handler, so an already-deleted Keycloak user threw on the retry path. #758 names exactly this as the trigger. Good catch.

## Blocking-ish 1 — the transaction is still batch-wide

One user is still able to take down the whole run. If any user in the batch raises a genuine `DataIntegrityViolationException`, Hibernate marks the transaction rollback-only and every other user's completed database work is discarded, while their Matrix and Keycloak side effects have already happened. The idempotency work in this PR reduces the *damage* of the replay; it does not stop the rollback.

This is not hypothetical. `user_chat` still has a `RESTRICT` FK to `user` (`chat_user_ibfk_1`) and **no delete step anywhere in the workflow** — 6 rows currently on PreDev. The first asker who is a group-chat participant will reintroduce exactly the loop this PR is meant to end.

Suggested shape: move the boundary to one transaction **per user** (`performUserDeletion` with `REQUIRES_NEW`, or a per-user method on the batch), so a poisoned user is isolated and the rest of the batch commits. Batch-level is only safe once the cascade is provably complete, and it is not.

## Blocking-ish 2 — the batch is not filtered to anonymous users

`deleteOverdueUsers()` selects on `findByStatus(DONE)` alone, with no `registration_type` filter, in a workflow named "anonymous" and configured under `user.anonymous.deleteworkflow.*`.

On PreDev this is currently selecting **registered system accounts**:

| session | user_id | registration_type | status |
| ---: | --- | --- | ---: |
| 103302 | `group-chat-system-1` | REGISTERED | 3 |
| 103430 | `group-chat-system-84` | REGISTERED | 3 |

These are the two rows the hourly job has been trying and failing to delete since 18/20 July. Had the delete ever succeeded, the job would have removed the group-chat system users for tenants 1 and 84. The rollback loop has been masking this. Please either filter to `RegistrationType.ANONYMOUS` or exclude system accounts explicitly, and say which in the PR body.

## Sequencing

#758 states the order as #746 → #757 → rebase #759. #757 is `CONFLICTING` and #759 is stacked on it, which is why this has not moved. This PR targets `pre-dev` directly and is green — that makes it the cheapest path. Recommend: merge #746 now for the immediate stop-the-bleeding, then bring this out of draft with the two changes above and retire #759 rather than untangling the #757 stack.

Issue #543, opened 2026-07-22
Make UserService replica safety and residual state coupling measurable


Storypapst:
Parent: https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/524

## Evidence

UserService is deployed as a Kubernetes Deployment and live-chat availability now uses Redis, but the process still contains replica-local Ehcache caches, active-view and Matrix token/sync/presence ConcurrentHashMaps, and multiple scheduled jobs. A Deployment is therefore not sufficient evidence that the service is operationally stateless.

## Scope

- Inventory every replica-local cache, map, scheduler, and legacy Rocket.Chat dependency.
- Classify each as performance-only, correctness-relevant, or duplicate-side-effect risk.
- Emit bounded runtime identity and replica-safety metrics without user/message identifiers.
- Add multi-replica tests proving requests can move between replicas without losing correctness.
- Add scheduler leader/idempotency proof and detect duplicate executions.
- Feed the ORISO management/refactoring dashboard with before/after evidence.

## Acceptance criteria

- [ ] Dashboard distinguishes workload type from verified stateless behavior.
- [ ] Every local state component has an owner, risk, migration/retention decision, and signal.
- [ ] Correctness-relevant state is externalized or the workload carries an explicit single-replica constraint and alert.
- [ ] Scheduled jobs are leader-safe or idempotent and duplicate executions are measurable.
- [ ] Multi-replica integration proof passes without sticky sessions.
- [ ] UserService can be marked stateless verified only after all critical checks are green.

Storypapst:
### Matrix browser-token bootstrap slice published

Live seven-day SigNoz traces isolated a frontend fan-out at the UserService boundary:

| Evidence | Result |
|---|---:|
| `/matrix/me/token` root requests | 122 |
| Matrix password-update PUTs | 122 |
| Matrix login POSTs | 124 |
| Root latency | p50 551.09 ms / p95 668.4 ms |

Frontend #456 remains the parent for token invalidation/refresh ownership. Child issue Frontend #791 now tracks the measured duplicate-bootstrap slice, and draft PR Frontend #792 implements subject-bound credential reuse, same-subject singleflight, cross-subject isolation, and forced refresh after explicit invalidation.

[...]

Issue #551, opened 2026-07-22
Delete session topics before asker sessions


Storypapst:
Parent: OpenResilienceInitiative/ORISO-UserService#531

## Why

The natural asker hard-delete retry removed the Keycloak identity but retained the UserService user, session 103442, and its `session_topic` row. The shared session-deletion action deletes other session dependencies but not topics, so the session and user deletes fail on the remaining foreign-key chain.

## What to build

- Delete all session-topic relations in the shared session-deletion action before deleting the session.
- Apply the same dependency cleanup to full asker deletion and single-session cleanup.
- Preserve workflow-error reporting when topic deletion fails.

## Acceptance

- [ ] A regression test fails without session-topic cleanup and passes with it.
- [ ] Session topics are deleted before the parent session.
- [ ] Both full-asker and single-session deletion paths use the repaired action.
- [ ] The full UserService unit suite and production package build pass.
- [ ] After reviewed regular PreDev deployment, the normal account scheduler removes the retained Blinky row/session and unblocks final consultant, agency, and tenant UI cleanup.

kiodreambau:
## Status check: the code fix is on `pre-dev`, the runtime criterion is blocked by a different defect

The session-topic deletion landed in commit `b3ba0a83` ("delete session topics before sessions"), in the shared `DeleteRoomsAndSessionAction.performSessionDeletion`, so both the full-asker and the single-session paths use it as this issue required.

The remaining acceptance box — "after reviewed regular PreDev deployment, the normal account scheduler removes the retained Blinky row/session" — cannot be ticked, and the reason is not in this issue's scope.

Every anonymous-deletion run is rolled back before commit: `DeleteUserAnonymousService.deleteInactiveAnonymousUsers()` is `@Transactional` around the whole batch including the error-notification mail, and that mail throws a 404 resolving a tenant template for technical tenant 0. The exception escapes the transaction proxy. Session 103302 has been deleted-and-restored roughly 200 times since 18 July.

So this fix, like #475, has been correct and discarded hourly.

## What has to happen before this can close

1. #746 merged and deployed.
2. One natural scheduler run.
3. Confirm the retained session and user rows are gone.

Note that after #746 the **session** will be removed but the **user** row will not — that is #859, a separate defect where the user delete merges a stale `sessions` collection. Verified locally: session gone, user retained.

Related: #866 (the remaining unclosed constraints across user, consultant and agency).

Issue #531, opened 2026-07-20
Soft-deleted asker leaves invisible enquiry that blocks tenant cleanup


Storypapst:
## Problem

A soft-deleted advice seeker can leave an agency enquiry in status `1`. The enquiry disappears from the consultant request UI, but the last-consultant deletion guard still counts it as open and returns `CONSULTANT_IS_THE_LAST_OF_AGENCY_AND_AGENCY_HAS_OPEN_ENQUIRIES`.

This creates a cleanup deadlock: no consultant can accept or close the hidden enquiry, while the last consultant, agency and tenant cannot be deleted.

## Reproduction

1. Create an agency enquiry (session 103442 in the disposable PreDev run).
2. Soft-delete the advice seeker; the 48-hour safeguard starts.
3. Open the remaining consultant's request list: the enquiry is absent.
4. Delete the remaining consultant: the backend returns the open-enquiry guard.

## Acceptance

- [ ] Soft deletion either keeps an actionable, privacy-safe cleanup representation or transitions/cancels the open enquiry atomically.
- [ ] Last-consultant deletion never depends on an enquiry that no authorised UI actor can resolve.
- [ ] Admin maps this reason to explicit user-facing guidance rather than leaving the generic confirmation dialog unchanged.
- [ ] Scheduler hard deletion removes or transitions the orphaned session and unblocks agency/tenant cleanup.
- [ ] Regression coverage includes the 48-hour safeguard and final hard-delete cascade.

## Evidence

- `24-consultant-3-delete-force-required-failed-predev-*`
- `24-consultant-3-cleanup-enquiry-103431-accepted-predev-*`
- `24-consultant-3-cleanup-enquiry-slot-1-open-requests-failed-predev-*`
- DB readback: session 103442 remains status 1 and unassigned while its user is `READ_ONLY_SAFEGUARD`.

Parent gate: OpenResilienceInitiative/ORISO-E2E#14

Storypapst:
## Cleanup deadlock evidence

The last-consultant deletion guard correctly reports an open agency enquiry, but the dialog cannot identify an actionable request:

![Last-consultant guard desktop](https://github.com/user-attachments/assets/c7dd2931-1e4f-4584-9397-4714f62201b4)

![Last-consultant guard mobile](https://github.com/user-attachments/assets/6c1ad120-7609-4faa-add5-87342f8a7318)

The visible enquiry was accepted successfully through the real App:

[...]

Issue #475, opened 2026-07-18
Anonymous-user cleanup fails: case_handover_request FK blocks session delete


Storypapst:
**Affected repos:** ORISO-UserService (possibly ORISO-Database for the FK migration)

## Why
The scheduled anonymous-user cleanup job aborts with `DataIntegrityViolationException` /
`ObjectOptimisticLockingFailureException`. A `case_handover_request` row references the
`session` being deleted through FK `case_handover_request_session_fk` (defined
`ON UPDATE CASCADE`, with no `ON DELETE` rule), so `delete from session where id=?` fails.
Consequence: anonymous users that have an associated case-handover request are **never
cleaned up**. This is a regression introduced by the Case Handover feature (CAR-CHO-01, ADR-002).

### Evidence (SigNoz, Pre-Dev, 12–15 Jul 2026)
- Service: `user-service`
- `org.springframework.dao.DataIntegrityViolationException` ×5 and
  `org.springframework.orm.ObjectOptimisticLockingFailureException` ×1 — **still failing**
  (latest run 2026-07-15 20:00, the deletion cron).
- Stacktrace: `DeleteAskerRoomsAndSessionsAction.execute:33`
  ← `DeleteUserAccountService.performUserDeletion:81`
  ← `DeleteUserAnonymousService.deleteInactiveAnonymousUsers:38`.
- Message: `Cannot delete or update a parent row: a foreign key constraint fails
  (userservice.case_handover_request, CONSTRAINT case_handover_request_session_fk
  FOREIGN KEY (session_id) REFERENCES session (id) ON UPDATE CASCADE) [delete from session where id=?]`

## What to build
- In the asker/session deletion action chain (`DeleteAskerRoomsAndSessionsAction` and related
  delete actions), delete or detach `case_handover_request` rows that reference the session
  **before** deleting the session.
- Decide the correct FK semantics: either add an explicit delete step for
  `case_handover_request`, or change the FK to `ON DELETE CASCADE` / `SET NULL` via a DB
  migration (coordinate with ORISO-Database — Liquibase is currently disabled there).
- Add a regression test: create a session with an associated `case_handover_request`, run the
  anonymous-user deletion, assert both rows are gone and no exception is thrown.

## Acceptance
- [ ] Anonymous-user cleanup deletes a session that has an associated `case_handover_request`
      without throwing.
- [ ] No orphaned `case_handover_request` rows remain after deletion.
[...]

PR #835, opened 2026-07-27
fix(notification): replay durable enquiry scheduler lease (#837)

Storypapst:
## Why

The hourly enquiry-notification job runs on every UserService replica. On current `pre-dev`, two concurrent instances both read the same open enquiries and both send the same mail batch. The historical stability PR #757 contains a broad, stale version of the fix and cannot be reviewed safely as one unit.

This PR replays only the independently reviewable scheduler lease on current Matrix-only `pre-dev`.

## What changed

- acquires a global, durable `scheduled_task_claim` before any session, AgencyService, consultant or MailService work;
- bounds the enquiry lease to 30 minutes, below the hourly schedule interval;
- serializes active and expired claims with a pessimistic MariaDB row lock;
- safely treats a concurrent first-insert deadlock/constraint conflict as a lost claim after reading the winner;
- adds Liquibase changeset `0076` and Hibernate schema validation for the claim table;
- adds the real MariaDB race contract to every required MariaDB workflow;
- records the red duplicate-mail reproduction, focused proof and remaining replica ceiling in the stability documentation.

## Architecture boundary

The target remains Matrix chat plus the ORISO frontend, an ORISO-controlled Element Call/MatrixRTC fork and LiveKit. Rocket.Chat and Jitsi are not supported fallbacks. UserService production source has neither runtime dependency; remaining Rocket.Chat names are removal migrations and regression contracts that prevent reintroduction.

This PR does **not** increase replicas, merge, deploy, or claim PreDev/product E2E proof. The broader #543 remains open for the other schedulers, authenticated writes and deployed telemetry.

## Validation

- red proof before implementation: two concurrent service instances sent 2 mail batches instead of 1;
- focused scheduler/enquiry suite: 15 tests, 0 failures, 0 errors;
- full unit suite: 3,382 tests, 0 failures, 0 errors, 4 environment-gated skips;
- full integration/contract/E2E suite: 841 tests, 0 failures, 0 errors, 4 environment-gated skips;
- fresh MariaDB 11.0.6: 102 Liquibase changesets plus 3 schema/statistics/lease contracts, green;
- MariaDB lease proof: concurrent initial insert and expired renewal each produced exactly one winner;
- Python CI and OpenAPI contracts: 21 tests, green;
- Spotless and `git diff --check`: clean;
- disposable MariaDB removed; no test processes or listeners retained.

## Relationship to the old stack

This supersedes only commit `a7e0ad1c` (enquiry scheduler lease) from #757, adjusted to the current changeset sequence and required MariaDB gate. The rest of #757 remains unchanged while its work is replayed in small current-`pre-dev` slices.

Closes OpenResilienceInitiative/ORISO-UserService#837
Refs #543
Refs #757

## Summary by CodeRabbit

* **New Features**
  * Added durable scheduling claims for enquiry notifications, ensuring only one application instance processes each notification batch.
  * Claims expire after 30 minutes by default and can be renewed when appropriate.
  * Added configurable claim duration and database support for storing claim information.

* **Bug Fixes**
  * Prevented duplicate notification processing when multiple application instances run concurrently.

* **Documentation**
  * Documented scheduler replica safety and claim behavior.


kiodreambau (commented):
Reviewed as the root of the five-deep lease chain (#835 → #836 → #840 → #842 → #844) while investigating the deletion-cascade failures (#475 / #531 / #551 / #745). The lease design itself is sound. Three things should change, and the chain shape is why none of this has moved.

## Design is right

`ScheduledTaskClaimWriter.claim` with `REQUIRES_NEW` + `findByTaskNameForUpdate` is the correct pattern: the claim commits independently of whatever the scheduled task later does, so a failing task cannot roll back its own claim. The `DataAccessException` → `hasActiveClaim` fallback correctly turns a unique-key race into `false` instead of an error. Durations are sensibly matched to the crons (PT30M for hourly jobs, PT12H for daily ones).

## 1 — Silent skip is the wrong default for deletion jobs

```java
if (!taskClaimService.tryClaim(TASK_NAME, claimDuration)) {
  return;
}
```

No log, no metric. This is exactly the failure-invisibility that made ORISO-AgencyService#138 undiagnosable for weeks — the report there literally reads "the job either never fires or finds no rows", because a silent scheduler is indistinguishable from an empty one.

These are GDPR/KDG deletion workflows. If a claim row is ever left with a far-future `claimedUntil` (clock skew between replicas, a mistyped duration, a manual insert during an incident), the deletion scheduler goes permanently quiet and nothing surfaces it. Please log at INFO on skip with the task name and the blocking `claimedUntil`, and consider a counter so a permanently-skipped deletion job is alertable.

## 2 — The claim is never released or extended

`claimedUntil` is set once, before the work runs, and there is no release on completion and no heartbeat. Effective semantics are therefore "run at most once per claim duration", not "one runner at a time". That is fine for the current crons, but it is worth stating in the class javadoc, because the name `tryClaim` reads like a mutex and someone will later shorten a duration below a task's runtime and get two concurrent runners.

Relevant context: PreDev is currently on deployment revision **404** of `oriso-platform-userservice` via "PreDev E2E hot deploy". Pods are replaced constantly. A pod killed mid-run leaves the claim held with no owner — with the current durations that costs one cycle, which is acceptable, but it argues for the skip log in point 1 so the lost cycle is visible.

## 3 — Chain shape is why this is stuck

Only #835 targets `pre-dev`; #836, #840, #842 and #844 are each based on their predecessor, all five are drafts, and a parallel `refactor/*` stack duplicates part of the same work. Nothing here can land until #835 lands, and #835 has had no review.

Suggestion: merge #835 on its own (it is self-contained — table, repository, writer, service, plus the enquiry-scheduler consumer), then retarget #836/#840/#842/#844 directly at `pre-dev` as four independent one-file PRs. They do not depend on each other, only on #835.

## Interaction with the deletion work

Note for whoever sequences this: #840 changes `DeleteUserAnonymousScheduler`, and #746/#824 change `DeleteUserAnonymousService` beneath it. They do not conflict textually, but #746 is the one currently stopping an hourly rollback loop on PreDev and should go first.

shazia-k (approved):

Issue #839, opened 2026-07-27
Lease anonymous user deletion scheduler across replicas


Storypapst:
## Why

The anonymous-user deletion scheduler runs on every replica and can duplicate database, Matrix, identity, and notification side effects for the same cleanup window.

## What to build

- Reuse the durable scheduler claim foundation before any deletion work.
- Keep the claim bounded below the schedule interval.
- Add red/green two-instance proof and preserve existing integration behavior.

## Acceptance

- [ ] The unfixed concurrent scheduler executes the workflow more than once.
- [ ] The fixed workflow has exactly one winner and a no-op loser.
- [ ] Existing scheduler integration cases clean claim state explicitly.
- [ ] Full unit, integration, contract, and CI gates are green.

Storypapst:
Verified as delivered: PR #840 (merged to pre-dev on 2026-08-07) closes #839: it adds a bounded durable lease claim to the anonymous-user deletion scheduler so exactly one replica runs the deletion workflow while other replicas are a clean no-op, adds a two-instance red/green regression test, and makes every existing scheduler integration test clear claim state explicitly. The full unit suite (3,386 tests) and the integration/contract suite (841 tests) passed with zero failures.

Moving to **QA external Dev** for external review. cc @BjoernLudwig

PR #840, opened 2026-07-27
fix(lifecycle): replay anonymous deletion lease (#839)

Storypapst:
## Why

The hourly anonymous-user deletion scheduler runs on every UserService replica. On the current code, two concurrent instances both set the technical tenant context and start the irreversible deletion workflow.

This PR replays only the anonymous-deletion lease from old #757. It is intentionally stacked on #836 so the durable claim primitive and the preceding lifecycle scheduler leases are not duplicated.

## What changed

- claims `anonymous-user-deletion` before setting tenant context or invoking the deletion service;
- bounds the claim to 30 minutes, below the hourly scheduler interval;
- makes a losing replica perform zero tenant-context, database, Matrix or identity deletion work;
- clears the durable claim around every existing scheduler integration test so test order cannot hide behavior;
- adds a two-instance red/green regression proof and records the contract in the stability document.

## Dependency and review shape

- **Base/dependency:** #836 (`replay/anonymous-deactivation-lease-543`)
- **Review delta:** six files, 134 added lines; no migration or generic claim code duplicated
- after #836 merges, this PR can be retargeted without changing its own commit

The target architecture remains Matrix chat plus the ORISO frontend, an ORISO-controlled Element Call/MatrixRTC fork and LiveKit. Rocket.Chat and Jitsi are not supported fallbacks.

## Validation

- red proof before implementation: two concurrent scheduler instances triggered the tenant context twice and the deletion workflow twice;
- focused unit/replica suite: 3 tests, green;
- focused Spring scheduler integration: 4 tests, green;
- full unit suite: 3,386 tests, 0 failures, 0 errors, 4 environment-gated skips;
- full required integration/contract/E2E suite: 79 reports, 841 tests, 0 failures, 0 errors, 4 environment-gated skips;
- Python CI/OpenAPI contracts: 21 tests, green;
- packaged application JAR, Spotless and `git diff --check`: green.

## Relationship to the old stack

This supersedes only commit `ed84507a` (anonymous deletion scheduler lease) from #757. Other scheduler work remains separate. Nothing is merged or deployed, and #543 remains open.

Closes OpenResilienceInitiative/ORISO-UserService#839
Refs #543
Refs #757
Depends on #836


kiodreambau (commented):
Reviewed as part of the deletion-cascade investigation (#475 / #531 / #551 / #745). The lease itself is fine; my substantive review of the design is on the chain root, #835 — the two points that apply here are the **silent skip** (`return` with no log, on a GDPR-relevant deletion job) and the fact that this PR is stacked three deep and cannot land until #835 does.

One PreDev-specific note: `user.anonymous.deleteworkflow.cron` is hourly (`0 0 * * * ?`), and the run currently issues roughly 252 Matrix calls. `PT30M` leaves comfortable headroom today, but the claim is taken before the work and never extended, so if that fan-out grows the margin shrinks silently. Worth a comment in the properties file recording the assumption.

Sequencing: #746 changes `DeleteUserAnonymousService` underneath this scheduler and is currently the fix that stops an hourly rollback loop on PreDev. Land that first, then retarget this at `pre-dev` directly once #835 is in.

shazia-k (approved):

Issue #872, opened 2026-07-28
[Bug]: Pre-Dev returns 403 for every authenticated request — tenant privacy timestamp fails OffsetDateTime parsing in HttpTenantFilter


kiodreambau:
### Affected Service/Component

User Service (consumer), Tenant Service (provider)

### Priority

P0-Blocker

### Problem

Since the Pre-Dev UserService rollout at **2026-07-27T23:10:47Z** (image built from `5eb17ab23b56223f26474cb4a19ead482aaa89dc`, PR #811 "clear all fixable HIGH/CRITICAL image vulnerabilities"), **every authenticated request for a tenant that has a privacy activation date returns HTTP 403**. The Pre-Dev app layer is effectively down for tenant 1 (`caritas-berlin`): consultants cannot log in, and no chat, live chat or session endpoint is reachable.

The failure is a provider/consumer date-format mismatch that the Spring Boot `4.0.1 -> 4.0.7` bump in that PR turned from latent into fatal:

- TenantService serializes `RestrictedTenantDTO.content.dataPrivacyConfirmation` from `TenantEntity.contentPrivacyActivationDate`, a `LocalDateTime`, so the wire value has **no offset**: `"2026-07-27T14:36:57"`.
- UserService's generated client declares the same field as `type: string, format: date-time` (`services/tenantadminservice.yaml:1165-1169`), which generates `OffsetDateTime`. The new Jackson 3 stack (`tools.jackson.*`) rejects the offset-less value.
- The call sits in `HttpTenantFilter.resolveSubdomain` (`HttpTenantFilter.java:78`), a servlet filter that runs for every non-whitelisted request. The exception escapes the filter chain before authorization completes, so Spring Security answers **403**.

Only `getSubdomain()` is used from that DTO — the date field is never read by UserService.

### Actual behavior

```
GET https://api.oriso-dev.site/service/users/data   (valid consultant bearer token)
HTTP 403
```

UserService log, once per request:

```
java.time.format.DateTimeParseException: Text '2026-07-27T14:36:57' could not be parsed at index 19
  at ...TenantService.getRestrictedTenantData(TenantService.java:33)
  at ...HttpTenantFilter.resolveSubdomain(HttpTenantFilter.java:78)
  at ...HttpTenantFilter.doFilterInternal(HttpTenantFilter.java:58)
Wrapped by: tools.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type
[...]

Issue #188, opened 2026-06-25
Consultant login fails while loading profile relations


Storypapst:
## Background
A newly created consultant can authenticate with Keycloak, but `GET /service/users/data` returns HTTP 500 while building the consultant login profile. The observed production correlation IDs include `bb4c9056-32fc-4a90-b5cc-a123e60730c3` and a direct API repro with `70e0a97f-20af-4b3c-838b-ace8596719eb`.

## Expected
Consultants with valid agency and language relations can log in and receive their user data profile.

## Actual
The consultant reaches the app error page after authentication. The UserService profile path accesses consultant relations after OSIV has been disabled, so lazy relation loading can fail during login profile assembly.

## Fix direction
Eagerly fetch the consultant agency/language login profile relations and keep the profile mapping read-only transactional.

## Validation
- Direct API repro confirmed Keycloak login succeeds and `/service/users/data` fails with HTTP 500 before the fix is deployed.
- Targeted local tests cover the profile provider and repository entity graph.

Storypapst:
Tracked in #189.

Storypapst:
Resolved via PR#189 (merged 2026-06-25). Closing as part of the GitHub hygiene sweep — issue was already fixed but never marked closed.

PR #869, opened 2026-07-28
fix(contracts): keep admin tenant timestamps as wire strings (#865)

Storypapst:
## Summary

- keep the TenantAdminService confirmation timestamps as plain wire strings because the provider emits offset-free local timestamps, not RFC 3339 date-times
- prove the administrative client deserializes the provider-shaped response without global generator type remapping
- ratchet only the administrative schema so `format: date-time` cannot return

## Why

Aggregate-only PreDev traces exposed successful TenantService responses failing inside the UserService client after transport. Parallel incident PR #874 owns the public-client/`HttpTenantFilter` outage path.

This PR owns the remaining administrative twin. The same two fields exist in `tenantadminservice.yaml` and receive the same offset-free provider values. Declaring them as RFC 3339 `date-time` would generate an incompatible temporal type even though UserService does not consume these timestamps as dates.

## Ownership reconciliation with #874

- #874 exclusively owns `services/tenantservice.yaml` and the public-client outage regression
- #869 exclusively owns `services/tenantadminservice.yaml` and its administrative behavioral/CI contracts
- an intermediate head accidentally reintroduced the public schema and test; exact head `fd178e2cec333b48ee2fdb2ef7cf2876979ac1d4` removes that overlap again
- `services/tenantservice.yaml` is byte-identical to `origin/pre-dev` on this head
- an exact synthetic merge of #874 head `dea5c01c6d097669b3dc09d44164a5b0de7dd3be` with this head is conflict-free

## Red-green proof

- before the revision, the administrative behavioral test received `LocalDateTime` instead of the contract-honest wire `String`
- the CI contract failed for both administrative confirmation fields while `format: date-time` remained
- after the revision, both tests pass and the original offset-free values are preserved as strings
- the later ownership regression produced a real merge conflict with #874; the revised exact head removes the duplicate public files and the conflict

## Verification on exact head

- clean unit suite: 3,374 tests, 0 failures/errors, 4 skips
- clean integration suite: 840 tests, 0 failures/errors, 3 environment-dependent skips
- Python CI contracts: 14 tests, green
- package build, Spotless and diff checks: green
- focused administrative Java and Python contract tests: green
- all 12 GitHub checks on the exact head are terminal green, including both
  full integration jobs, branch and PR builds, Docker validation, Redis, both
  MariaDB jobs, OpenAPI contracts and Required PreDev CI

No database reset was needed; integration tests used isolated test schemas.

## Runtime acceptance after deployment

[...]

PR #874, opened 2026-07-28
fix(tenant): stop 403-ing every request on a local-date-time privacy stamp (#872)

kiodreambau:
## Why

Pre-Dev has been answering **HTTP 403 to every authenticated request for tenant 1** since the UserService rollout at `2026-07-27T23:10:47Z`. No consultant can log in, so nothing behind authentication — chat, live chat, sessions — is reachable. Found while running the P3 live-chat E2E gate, which is blocked at step 1.

`TenantService` serializes `RestrictedTenantDTO.content.dataPrivacyConfirmation` from `TenantEntity.contentPrivacyActivationDate`, a `LocalDateTime`, so the wire value carries no UTC offset:

```
GET /service/tenant/public/id/1  ->  200
"dataPrivacyConfirmation": "2026-07-27T14:36:57"
```

Our client spec declared that field `format: date-time`, which generates `OffsetDateTime`. The Jackson 3 stack that came with the Spring Boot `4.0.1 -> 4.0.7` bump in #811 rejects the offset-less value:

```
java.time.format.DateTimeParseException: Text '2026-07-27T14:36:57' could not be parsed at index 19
  at ...TenantService.getRestrictedTenantData(TenantService.java:33)
  at ...HttpTenantFilter.resolveSubdomain(HttpTenantFilter.java:78)
Wrapped by: tools.jackson.databind.exc.InvalidFormatException: Cannot deserialize value of type
  `java.time.OffsetDateTime` from String "2026-07-27T14:36:57"
```

`HttpTenantFilter` runs for every non-whitelisted request and the exception escapes the filter chain before authorization completes, so Spring Security answers 403 for the whole tenant.

## What this changes

`services/tenantservice.yaml`: `dataPrivacyConfirmation` and `termsAndConditionsConfirmation` stay plain strings instead of `format: date-time`. UserService only reads `getSubdomain()` from this DTO — neither timestamp is used anywhere — so keeping them untyped removes the failure mode without touching the provider contract that Admin and Frontend also consume.

`termsAndConditionsConfirmation` had the identical defect and would have fired as soon as a tenant set a terms activation date.

## Verification

- **Red:** with the spec change reverted, `RestrictedTenantDataContractTest` does not compile — `incompatible types: OffsetDateTime cannot be converted to String`. That is the intended tripwire if `format: date-time` ever returns.
- **Green:** `./mvnw -Dtest=RestrictedTenantDataContractTest test` -> `Tests run: 2, Failures: 0, Errors: 0`.
- Spotless clean.
- The test deserializes the verbatim Pre-Dev payload through `MappingJackson2HttpMessageConverter`, the same converter the generated client uses.

## Deployment note

This needs a Pre-Dev deploy to unblock the environment — Pre-Dev is down for tenant 1 until then. Affected tenants: `1 caritas-berlin` (set 2026-07-27 14:36:57) and `24 Test Mandant` (set 2026-05-13 10:20:41).

## Worth a look separately

The Spring Boot 4.0.7 / Jackson 3 bump may have made other cross-service `format: date-time` fields strict in the same way. The approved-but-unmerged contract PRs (TenantService #137, AgencyService #188) look related.

Closes OpenResilienceInitiative/ORISO-UserService#872

### DIFF
--- a/services/tenantservice.yaml
+++ b/services/tenantservice.yaml
@@ -1159,14 +1159,22 @@ components:
           example: Llorem ipsum...
         termsAndConditionsConfirmation:
           type: string
-          format: date-time
           example: '2022-12-07T11:51:01'
-          description: Datetime stamp when new terms and condition defined
+          description: >-
+            Datetime stamp when new terms and conditions were defined.
+            TenantService serializes this from a `LocalDateTime`
+            (`TenantEntity.contentTermsAndConditionsActivationDate`), so the wire
+            value carries no UTC offset and is not an RFC 3339 date-time. It must
+            stay a plain string here: `format: date-time` generates an
+            `OffsetDateTime` whose deserialization fails and takes down every
+            request through `HttpTenantFilter`.
         dataPrivacyConfirmation:
           type: string
-          format: date-time
           example: '2022-12-07T11:51:01'
-          description: Datetime stamp when new privacy defined
+          description: >-
+            Datetime stamp when a new privacy text was defined. Same
+            local-date-time wire format as `termsAndConditionsConfirmation` —
+            see the note there.
         dataProtectionContactTemplate:
           $ref: '#/components/schemas/DataProtectionContactTemplateDTO'
     Settings:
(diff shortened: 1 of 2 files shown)


Storypapst:
Cross-PR reconciliation completed with #869.

A temporary checkout combining exact head `dea5c01c6` from this PR with the previous #869 head passed both behavioral regressions and the CI contract. #869 has now been narrowed to the non-overlapping administrative schema only: it removes its public-client generator mapping and preserves the same offset-free fields as plain wire strings in `tenantadminservice.yaml`.

Resulting ownership:
- #874: public TenantService schema and the urgent `HttpTenantFilter` outage path
- #869: administrative TenantService schema parity

Both still require normal review, merge, deployment, and aggregate-only runtime acceptance; neither is being represented as live yet.

shazia-k (approved):

Issue #890, opened 2026-07-28
[TEN-INV-U6] Correct invite link targets and SENT only after confirmed delivery


kiodreambau:
**Priority:** Urgent *(Roadmap Priority field has no options yet — see parent)*

## Why
Tenant-admin invite links point to the App layer although the flow must complete on a public Admin page, and SENT is set without proven delivery. The tenant is an organisation, not an app user — the login belongs to the tenant admin.

## What to build
- Tenant-admin invite links target the public Admin route; counsellor invite links keep the public App route.
- `SENT` and the success message are set only after the mail service (U5) confirms delivery; on failure no false success appears.
- Consumed, revoked or expired links return a distinct error and cannot be reused.

## Acceptance
- [ ] A real email arrives and contains the correct Admin or App link respectively.
- [ ] SMTP failure produces neither SENT nor a success toast.
- [ ] A consumed/revoked/expired link shows a clear error.

Part of OpenResilienceInitiative/ORISO-Admin#569

## Hard evidence from PreDev QA (2026-07-28) — MUST be covered by this chunk

- **No mail is ever dispatched:** `AccountInviteService.sendInvite()` (~line 260-293) renders subject/body into an `invite_email_delivery` row with `status = SENT` hard-coded. There is **no MailService call anywhere in the account-invite package** and no consumer of `InviteEmailDeliveryRepository`. The FAILED path is unreachable code. Fix = real dispatch + status from the actual transport outcome.
- **`acceptBaseUrl` is trusted verbatim from the client** (`buildAcceptUrl()`): any authorized API caller can mint invite links to an arbitrary domain — phishing vector the moment mail actually sends. Derive the base URL server-side from role/config (TENANT_ADMIN -> public Admin base URL, COUNSELLOR -> App base URL) or enforce a strict origin allowlist; do not keep the free-form client parameter.
- **Link targets today are dead ends:** Admin hardcodes `appURL + /account-invite`; the App (ORISO-Frontend) has **zero** `account-invite` routes (the public counsellor acceptance route is OpenResilienceInitiative/ORISO-Frontend#520); anonymous open silently redirects to the advice-seeker login.

Evidence: CSV QA run 2026-07-28, captures in the session scratchpad (`csv-test/`), e.g. `A-posts.json` (201 + EMAIL_SENT/SENT with no mail), `07-acceptUrl-app-layer.png`.

Storypapst:
[...]

Issue #889, opened 2026-07-28
[TEN-INV-U3] Coordinate tenant/agency ID reservations during invite creation


kiodreambau:
**Priority:** Urgent *(Roadmap Priority field has no options yet — see parent)*

## Why
Invite creation must orchestrate the U1/U2 reservations so that parallel invites can never end up with the same ID and failed creations leave no orphaned reservations.

## What to build
- On invite creation, reserve the tenant (and agency, per U2 decision) ID via the authoritative service operations and re-validate immediately before saving.
- Reject concurrent conflicts: the second request for the same ID receives a conflict response.
- Compensation: if creation fails after reservation, release the reservation — no orphans.
- Provide the Admin UI with an aggregated live-validation endpoint for tenant and agency IDs (states: free, reserved by invite, already assigned).

## Acceptance
- [ ] Two parallel invites in AUTO mode get different IDs.
- [ ] Two parallel manual invites for the same ID: exactly one succeeds, the other gets a conflict.
- [ ] A failed invite creation leaves no reservation behind.
- [ ] The green UI state alone grants nothing — the server check on submit is authoritative.

Part of OpenResilienceInitiative/ORISO-Admin#569

Storypapst:
Verified as delivered: PR #906 (merged into pre-dev on 2026-07-29) implements authoritative tenant/agency ID reservation on invite creation, re-validation before save, best-effort compensation on failure, 409 conflict handling for concurrent same-ID requests, and the new `GET /useradmin/id-allocation` live-validation endpoint. All four acceptance criteria are directly exercised by `AccountInviteReservationOrchestrationIT` plus dedicated controller/service tests.

Moving to **QA external Dev** for external review. cc @BjoernLudwig

PR #907, opened 2026-07-29
[TEN-INV-U6] Correct invite link targets and SENT only after confirmed delivery

kiodreambau:
**Stack position: 2 of 6. Base: `fix/ten-inv-u3-invite-reservation-orchestration` (PR 1). Review only the top commit.**

## Problem

Three defects confirmed by the PreDev QA run of 2026-07-28, all recorded on
OpenResilienceInitiative/ORISO-UserService#890:

1. **No mail was ever dispatched.** `AccountInviteService.sendInvite()` rendered subject and
   body into an `invite_email_delivery` row with `status = SENT` hard-coded. There was no
   MailService call anywhere in the account-invite package and no consumer of
   `InviteEmailDeliveryRepository` — the `FAILED` path was unreachable code. The API
   answered `201` with `EMAIL_SENT`/`SENT` while nothing left the building.
2. **`acceptBaseUrl` was trusted verbatim from the client.** Any authorized API caller could
   mint invite links pointing at an arbitrary domain — a phishing vector the moment mail
   actually started sending.
3. **Link targets were dead ends.** Tenant-admin links pointed at the App layer, which has no
   `account-invite` route, so an anonymous open silently redirected to the advice-seeker login.
   Consumed, revoked and expired links were also indistinguishable to the client.

## What changed

**SENT only after confirmed delivery**

- New `InviteMailDispatchService` + `InviteMailTransport` with a strict
  receipt-after-send contract (mirroring CTS TEN-INV-U5): either the SMTP server accepted the
  message and a receipt comes back, or `SmtpSendException` propagates as
  `502 SMTP_SEND_FAILED`. Connection settings come from the public CTS `/settings` payload;
  credentials from `smtp.user`/`smtp.password` or the super-admin credentials endpoint.
- `sendInvite()` now dispatches **before** persisting. `EMAIL_SENT` and the `SENT` delivery
  row are written only after the transport receipt; a failure rolls the send back and leaves
  at most a `FAILED` audit row (`InviteEmailDeliveryFailureRecorder`, `REQUIRES_NEW`).
- Resend supersedes the old invite only after the replacement mail was accepted — a failed
  handover leaves the old invite intact and usable.

**Server-side link targets**

- New `InviteAcceptUrlBuilder` derives the accept URL from role plus configuration:
  `TENANT_ADMIN` → `<admin-base>/admin/tenant-onboarding/{token}` (the Admin U8 public route),
  every other role → `<app-base>/account-invite/{token}`.
[...]

PR #906, opened 2026-07-29
[TEN-INV-U3] Coordinate tenant/agency ID reservations during invite creation

kiodreambau:
**Stack position: 1 of 6 — bottom of the chain. Base: `pre-dev`.**

## Problem

Tenant-admin invite creation allocated no ID authoritatively. Two invites created in
parallel could end up pointing at the same tenant ID, a failed creation left the ID
half-claimed, and the Admin UI had no server-backed way to tell an operator whether a
tenant or agency ID was actually free before submitting. The "green" client-side state
was the only signal, and it granted nothing.

## What changed

- **Authoritative tenant-ID reservation.** Invite creation reserves the tenant ID in
  TenantService (`POST /tenantadmin/tenant-ids/reservations`, `AUTO` or a pinned ID) and
  stores the reservation token on the invite (new column
  `account_invite.tenant_id_reservation_token`, changeset `0076`), so tenant creation can
  later consume exactly the reserved ID.
- **Agency IDs** are reserved in AgencyService's own ID space (per the U2 decision,
  ORISO-AgencyService#214) when the composer sends an agency allocation mode; tenant IDs
  are only validated there, never minted.
- **Re-validation before save.** The `RESERVED` state is re-checked immediately before the
  invite row is written; a lost race maps to `409` and releases the reservation.
- **Compensation.** Any creation failure after reservation releases both the tenant and the
  agency reservation, so no orphans accumulate. Release is best-effort and never masks the
  original failure.
- **Concurrent conflicts.** A second manual invite for the same ID gets `409` from the
  upstream ledger, passed through as `ConflictException`.
- **Resend** carries the reservation token over to the replacement invite.
- **New live-validation endpoint for the Admin UI:**
  `GET /useradmin/id-allocation?tenantId=&agencyId=` returning per-ID
  `FREE` / `RESERVED` / `ASSIGNED`, or `SERVICE_ERROR` with upstream status passthrough.
- `services/tenantadminservice.yaml` and `services/agencyadminservice.yaml` extended with
  the real U1/U2 endpoint contracts so the generated clients pick them up.

The submit-time server check stays authoritative: the aggregated endpoint is advisory UI
sugar, it does not reserve anything.

## Testing

Re-run locally on 2026-07-29 with JDK 21 (`./mvnw test` / `surefire:test@integration-tests`,
targeted classes) on the tip of this stack — all six branches applied:

| Suite | Result |
| --- | --- |
| `AccountInviteServiceTest` | 83 tests, 0 failures |
[...]

PR #908, opened 2026-07-29
[TEN-INV] Invite hardening: strict STARTTLS, atomic accept claim, resumable 2FA onboarding

kiodreambau:
**Stack position: 3 of 6. Base: `fix/ten-inv-u6-invite-links-and-sent-status` (PR 2). Review only the top commit.**

This is the review-hardening PR for the two chunks below it. It closes findings raised in the
U3/U6 verify pass rather than delivering a sub-issue of its own, so it uses `Refs`, not `Closes`.

## Problem

Findings from the U3/U6 verify pass, plus one latent crash:

- **Credentials could go out in plaintext.** The SMTP transport enabled STARTTLS but did not
  *require* it. A server — or a MITM — that simply does not advertise STARTTLS caused a silent
  fallback to an unencrypted session carrying the operator's SMTP credentials. Server identity
  was also not checked against the certificate.
- **Single-use accept depended on a lock hint.** The accept path relied on a
  `PESSIMISTIC_WRITE` hint on the token lookup to serialize two concurrent accepts. That hint
  happens to be honoured by H2, but it is not an invariant to build single-use semantics on.
- **Tenant admins could get stranded mid-onboarding.** A tenant admin who completed account
  creation but dropped out before activating TOTP came back to a terminal `410 CONSUMED` — the
  link was spent and the account unusable, with no way forward.
- **Deployment order produced an unmapped 500.** `createInvite` for `TENANT_ADMIN` crashed with
  an unmapped 500 when TenantService did not yet expose the TEN-INV-U1 allocation endpoints (404).
- **Latent NPE:** creating any invite without a `tenantId` unboxed `command.tenantId()` through
  the reservation ternary (`TenantIdReservation` holds a primitive `long`) and crashed 500.

## What changed

**SMTP transport (`JakartaInviteMailTransport`)**

- `mail.smtp.starttls.required=true` in STARTTLS mode — a server that does not offer STARTTLS
  is refused instead of silently receiving the credentials in the clear.
- `mail.smtp.ssl.checkserveridentity=true` in both modes; the certificate must match the
  configured host.

**Atomic single-use accept**

- `AccountInviteRepository.claimForAcceptance()` flips `EMAIL_SENT → ACCEPTED` as one guarded
  `UPDATE`. Of two concurrent accepts exactly one sees an affected row, independent of whether
  the pessimistic lock hint is honoured. The loser re-reads the winner's committed state and
  runs through the normal terminal/resume logic.

**Resume contract for stranded tenant-admin onboarding**

- `TENANT_ADMIN` invites now default to `twoFactorStatus=PENDING_SETUP` (tenant-admin TOTP is
  mandatory per ORISO-Admin#569).
- A consumed invite whose 2FA gate is still `PENDING_SETUP` stays resumable inside the original
  expiry window: `POST /users/account-invites/{token}/accept` answers an idempotent `200` with
  `phase=PENDING_2FA_ACTIVATION` — same response shape and data as the original accept, the
  `ACCEPTED` audit state untouched — instead of a terminal `410 CONSUMED`. Once the gate is
  satisfied (`ACTIVE` via the OTP lifecycle, `WAIVED`, `NOT_REQUIRED`) or the link expires, the
  link is terminally consumed (`410 reason=CONSUMED`).
- New `AcceptPhase` enum and `phase` field on the accept response; `null` on admin-facing endpoints.

**Deployment-order degradation**

- Legacy requests without an allocation mode fall back to the pre-U3 duplicate checks with a
  warning when TenantService lacks the U1 endpoints. Explicit `AUTO`/`MANUAL` requests still fail
  loudly, now with a clear deployment-order message instead of an unmapped 500.

## Testing

Re-run locally on 2026-07-29 with JDK 21 on the tip of this stack:

| Suite | Result |
| --- | --- |
| `AccountInviteServiceTest` | 83 tests, 0 failures |
| `AccountInviteControllerTest` | 27 tests, 0 failures |
| `JakartaInviteMailTransportTest` | 4 tests, 0 failures |
| `AccountInviteAcceptRaceIT` | 2 tests, 0 failures |

`AccountInviteAcceptRaceIT` is new and proves the single-use invariant with two latched threads
against the real JPA layer. Worth knowing while reviewing: the pre-existing `PESSIMISTIC_WRITE`
hint already serialized this race on H2 — the guarded `UPDATE` removes the *dependency* on that
hint, it does not fix an observed H2 failure. That is exactly why it is worth having on MariaDB.

Chain-run evidence from the clean-slate local stack:

- **S5** — a 2FA-pending link resumed rather than dead-ending; a completed link reported `CONSUMED`.
- **S1** — the STARTTLS catcher accepted the invite mail with the required-STARTTLS transport.

**Honest note:** `UserControllerE2EIT` fails on this machine with `RedisConnectionFailureException`
(`Connection refused: localhost/127.0.0.1:6379`) — no local Redis, identical on the untouched
baseline, not a regression from this stack. No green full IT suite is being claimed.

Refs OpenResilienceInitiative/ORISO-UserService#889
Refs OpenResilienceInitiative/ORISO-UserService#890

Part of OpenResilienceInitiative/ORISO-Admin#569

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/AccountInviteController.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/AccountInviteController.java
@@ -17,6 +17,7 @@ import de.caritas.cob.userservice.api.service.accountinvite.InviteEmailDeliveryS
 import de.caritas.cob.userservice.api.service.accountinvite.InviteEmailTemplateKind;
 import de.caritas.cob.userservice.api.service.accountinvite.InviteEmailTemplateService;
 import de.caritas.cob.userservice.api.service.accountinvite.InviteEmailTemplateService.TemplateCommand;
+import de.caritas.cob.userservice.api.service.accountinvite.TwoFactorGateStatus;
 import de.caritas.cob.userservice.api.service.accountinvite.allocation.IdAllocationMode;
 import java.time.LocalDateTime;
 import java.util.List;
@@ -161,16 +162,25 @@ public class AccountInviteController {
             accountInviteService.calculateAccessGate(invite)));
   }
 
+  /**
+   * Public accept endpoint. Resume contract (ORISO-Admin#569 hardening): while the accepted
+   * invite's mandatory 2FA activation is pending and the link is unexpired, repeated calls stay
+   * idempotent 200s carrying {@code phase = PENDING_2FA_ACTIVATION}; once the gate is satisfied the
+   * link is terminally consumed (410 {@code reason=CONSUMED}).
+   */
   @PostMapping("/users/account-invites/{token}/accept")
   public ResponseEntity<AccountInviteResponseDTO> acceptInvite(
       @PathVariable String token, @RequestBody(required = false) AcceptInviteRequestDTO request) {
     String acceptedByUserId = request == null ? null : request.acceptedByUserId;
     AccountInvite invite = accountInviteService.acceptInvite(token, acceptedByUserId);
-    return ResponseEntity.ok(
+    AccountInviteResponseDTO response =
         AccountInviteResponseDTO.from(
-            invite,
-            latestDeliveryStatus(invite),
-            accountInviteService.calculateAccessGate(invite)));
+            invite, latestDeliveryStatus(invite), accountInviteService.calculateAccessGate(invite));
+    response.phase =
+        invite.getTwoFactorStatus() == TwoFactorGateStatus.PENDING_SETUP
+            ? AcceptPhase.PENDING_2FA_ACTIVATION.name()
+            : AcceptPhase.COMPLETED.name();
+    return ResponseEntity.ok(response);
   }
 
   @PreAuthorize(ADMIN_AUTH)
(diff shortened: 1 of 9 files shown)


kiodreambau:
### Stack order (ORISO-UserService, bug ORISO-Admin#569)

Reviewable stacked chain — each PR carries exactly one commit, each is based on the one before it.
**Merge strictly bottom-up.** After each merge GitHub retargets the next PR automatically; if it
does not, retarget it manually before merging.

| # | PR | Head | Base |
| --- | --- | --- | --- |
| 1 | #906 | `fix/ten-inv-u3-invite-reservation-orchestration` | `pre-dev` |
| 2 | #907 | `fix/ten-inv-u6-invite-links-and-sent-status` | #906 |
| 3 | #908 | `fix/ten-inv-hardening-us` | #907 |
| 4 | #909 | `fix/ten-inv-chainfix-userservice` | #908 |
| 5 | #910 | `fix/ten-inv-chainfix2-userservice` | #909 |
| 6 | #911 | `fix/ten-inv-dpa-continuity` | #910 |

**This PR is 3 of 6. Merge after #907.** Review hardening for the two chunks below it; single commit.

The head of #911 is byte-identical to the proven local integration branch `avv-fix`
(`git diff --stat fix/ten-inv-dpa-continuity..avv-fix` is empty, same SHA `1910ff3e`), so the union
of this stack is exactly the state that passed the clean-slate chain test — nothing was left out
and nothing extra was added.

No history was rewritten and no branch was force-pushed; every head points at the branch that was
already pushed to origin.

Issue #820, opened 2026-07-27
Refactor dummy-email updates behind a focused identity port


Storypapst:
## Context

The current `pre-dev` baseline still exposed registration-time dummy-email replacement through the broad `IdentityClient` and passed the web-layer `UserDTO` into the identity adapter. The earlier stacked implementation was based on an older source graph: its asker-import consumer no longer exists on the current Matrix-only baseline.

## Scope

- [x] Add a focused, provider-neutral dummy-email output port and value object.
- [x] Route the one remaining external consumer, user registration, through the focused port.
- [x] Remove dummy-email methods from the broad `IdentityClient`.
- [x] Keep current-user email deletion on the existing email-address operation.
- [x] Preserve the identity-provider call bound: one target resolution and one update.
- [x] Update shared Spring test wiring, focused unit tests, architecture contracts, and stability documentation.
- [x] Prove focused tests, the complete unit and required integration suites, Redis and fresh MariaDB contracts, packaging, formatting, and review.
- [x] Publish a direct ready-for-review PR against the current `pre-dev` baseline.

## Architecture boundary

ORISO is moving to Matrix chat through its own frontend, an ORISO-controlled MatrixRTC / Element Call fork, and LiveKit. Rocket.Chat and Jitsi are removal targets, not fallbacks. This identity refactor preserves that current Matrix-only source graph and does not restore removed Rocket.Chat, Jitsi, or asker-import paths.

## Out of scope

- Identity account creation, the remaining broad-client responsibility after this slice.
- Merge, deployment, and deployed runtime proof.
- Unrelated database or schema changes.

Parent: #335

Implementation: #900

Storypapst:
Local implementation and its direct CI are complete, but integration is blocked by stack drift: the root stability PR #757 conflicts with current `pre-dev`, which now contains merged Matrix-only removal PR #756. Before #821 can move to review, replay the full stack on current `pre-dev`, drop changes to deleted legacy asker-import/Rocket.Chat paths, and rerun all gates. No branch history has been rewritten yet.

Storypapst:
Implementation and verification are complete on draft PR #821: all 11 GitHub checks and the full local test suite are green; no database reset was needed.

[...]

Issue #791, opened 2026-07-27
Split username availability from the broad identity client


Storypapst:
Parent: #335

## Why

`IdentityClient` mixes account mutations with the read-only username-availability check. On current `pre-dev`, four production classes depend on that broad command surface for the read, while `UserVerifier` retains an unused broad-client dependency and commented legacy branch.

## Current source truth

The four current direct availability consumers are:

1. `UserController`
2. `UserRegistrationControllerDelegate`
3. `AnonymousUsernameRegistry`
4. `ConsultantImportService`

The older stacked implementation referenced an `IdentityManager`/`AskerImportService` structure that is no longer present. Removed services and legacy paths must not be restored.

## What to build

- Add a focused provider-neutral output port for username availability.
- Move all four current production availability consumers to the focused port.
- Remove the unused broad-client dependency and commented legacy branch from `UserVerifier`.
- Remove `isUsernameAvailable` from the broad `IdentityClient`.
- Preserve raw and encoded usernames and the public fail-open outage behavior.
- Add executable module-boundary and focused interaction contracts.
- Record unchanged outbound call behavior and release truth.

## Acceptance

- [x] Replacement is directly based on current `pre-dev`, not a stale stacked branch.
- [x] `IdentityClient` no longer exposes username availability.
- [x] All four current production consumers depend on the focused port.
- [x] Removed services and legacy paths are not restored.
- [x] Raw/encoded username behavior and public fail-open behavior are preserved.
- [x] One availability operation preserves exactly two adapter-internal identity searches: decoded and encoded username.
- [x] Java 21 unit, integration, architecture and contract suites remain green locally with zero quarantine.
- [x] Stability documentation records the boundary and unchanged two-search behavior.
- [x] Replacement PR GitHub CI is fully green and review-ready.

Implementation: #881
Supersedes: #792

This is modular-monolith boundary work, not a microservice extraction or a call-count reduction. The Matrix-only target remains unchanged: ORISO-owned frontend, ORISO-controlled Element Call/MatrixRTC fork and LiveKit. Rocket.Chat and Jitsi are complete removal targets, never fallbacks.

Storypapst:
[...]

PR #931, opened 2026-07-30
fix(test): restore every identity port on the shared keycloakService mock

shazia-k:
Fixes the `required integration tests` failure on `pre-dev` ([run 30542537111](https://github.com/OpenResilienceInitiative/ORISO-UserService/actions/runs/30542537111/job/90870504328), 157 errors).

## Root cause

`#900` added `IdentityDummyEmailUpdater` to the shared `@MockitoBean` on `IdentityClient`, but **replaced** `extraInterfaces` instead of appending to it:

```java
- @MockitoBean(
-     extraInterfaces = {
-       IdentityAuthentication.class,
-       IdentityEmailOwnerLookup.class,
-       IdentityUsernameAvailability.class
-     })
+ @MockitoBean(extraInterfaces = IdentityDummyEmailUpdater.class)
  private IdentityClient identityClient;
```

That mock replaces the entire `keycloakService` bean, and `KeycloakService` is the single implementation of all of those ports. Without them the ApplicationContext cannot be built for any test whose graph injects one, so the suite fails far from the cause:

```
BeanNotOfRequiredTypeException: Bean named 'keycloakService' is expected to be of type
'IdentityAuthentication' but was actually of type 'IdentityClient$MockitoMock$...'
  ... creating bean 'appointmentService', constructor parameter 5
```

Six classes were truncated: `UpdateAdminServiceIT`, `CreateAdminServiceIT`, `UserControllerAuthorizationIT`, `UserControllerIT`, `UserAdminControllerMultiTenancyTrueE2EIT`, `ConsultantAgencyRelationCreatorServiceTenantAwareIT`. Two (`ConsultantUpdateServiceBase`, `UserAdminControllerE2EIT`) were updated correctly and are untouched here.

## Changes

1. Restore the full port list in the six affected classes.
2. Add `test_shared_spring_identity_mocks_carry_every_keycloak_port` to `tests/ci/test_module_boundaries.py`, deriving the required set from `KeycloakService`'s `implements` clause rather than hardcoding it.

The existing guard only asserted `IdentityDummyEmailUpdater`, which is why `#900` passed `validate`. Splitting `IdentityClient` into focused ports has caused this exact breakage twice before (`#791`, `#820`); the new guard runs in the cheap `validate` job and names the missing ports, so the next split fails there instead of hundreds of errors deep in the integration suite.

## Verification

- `spotless:check` clean.
- All 56 CI contract guards pass (`python3 -m unittest discover -s tests/ci -p 'test_*.py'`).
- Confirmed the new guard **fails** with the exact missing-port list when one file is reverted to the broken state.
- Ran `scripts/ci/run-required-integration-tests.sh` locally: all six previously-failing classes now load and pass. The only remaining local failure is `UserControllerE2EIT`, whose 25 errors are all `RedisConnectionFailureException: Connection refused localhost:6379` — this machine has no Redis; the CI job provides a `redis:7-alpine` service, so that class is expected to pass there.

## Follow-up

`#899` (`replay/focused-identity-account-removal-818-direct`) adds a fifth port, `IdentityAccountRemover`, and branched before the others existed. Its conflict resolution must list **all** ports in `extraInterfaces` — the new guard will enforce this.

### DIFF
--- a/.github/actions/maven-build/action.yml
+++ b/.github/actions/maven-build/action.yml
@@ -38,6 +38,19 @@ runs:
           sys.exit(1)
       PY
 
+  # Also guards against a vacuously green run: the failure check above passes
+  # when no report exists at all, --require does not.
+  - name: Report unit inventory
+    if: ${{ always() }}
+    shell: bash
+    run: |
+      set -o pipefail
+      {
+        echo "### Unit test inventory"
+        echo
+        python3 scripts/ci/suite-inventory.py --require unit
+      } | tee -a "$GITHUB_STEP_SUMMARY"
+
   - name: Verify CI and load-test contracts
     shell: bash
     run: python3 -m unittest discover -s tests/ci -p 'test_*.py'
--- a/.github/actions/maven-verify-required/action.yml
+++ b/.github/actions/maven-verify-required/action.yml
@@ -18,6 +18,17 @@ runs:
     shell: bash
     run: scripts/ci/run-required-integration-tests.sh
 
+  - name: Report integration inventory
+    if: ${{ always() }}
+    shell: bash
+    run: |
+      set -o pipefail
+      {
+        echo "### Integration test inventory"
+        echo
+        python3 scripts/ci/suite-inventory.py --require integration
+      } | tee -a "$GITHUB_STEP_SUMMARY"
+
   - name: Upload required integration reports
     if: ${{ always() }}
     uses: actions/upload-artifact@v4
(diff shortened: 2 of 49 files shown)


shazia-k:
The conflict was two contract tests appended at the same point in
test_module_boundaries.py — this branch's generic guard vs
test_identity_account_removal_consumers_use_a_focused_port, which came in with
the IdentityAccountRemover split. Kept both.

PR #936, opened 2026-07-30
fix(logs): scope the admin audit logs to the agencies an admin owns

Storypapst:
## Problem

`/users/supervisors/logs` and `/users/case-handover/logs` filter by tenant only. Beratungsstellen-Admins (`restricted-agency-admin`, always paired with `user-admin`) already pass the authority check on both endpoints, so they can read the whole Träger's audit trail. The Admin panel is about to surface these views for that role (OpenResilienceInitiative/ORISO-Admin#84), which turns a latent gap into a visible one.

## What changed

- new `AdminAuditAgencyScope` resolves how far an admin may read: empty `Optional` for tenant-wide admins (platform, Träger, full `agency-admin`), the assigned agency set for a Beratungsstellen-Admin
- `SupervisorLogsService` and `CaseHandoverLogsService` append `s.agency_id IN (:agencyIds)` when a scope is present; tenant-wide admins keep the exact query they had
- fail closed: an admin with no agency assignment reads nothing and the database is never queried, instead of falling back to the tenant
- no `SecurityConfig` change — the authority set is already correct, only the data scope was too wide

## Verification

- `./mvnw test -Dtest='SupervisorLogsServiceTest,CaseHandoverLogsServiceTest,AdminAuditAgencyScopeTest'` — 31 tests, green (run on JDK 21; Mockito cannot mock under JDK 25 locally, which also fails the pre-existing tests)
- `./mvnw spotless:apply` clean

Refs OpenResilienceInitiative/ORISO-Admin#84

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/service/AdminAuditAgencyScope.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/AdminAuditAgencyScope.java
@@ -1,0 +1,49 @@
+package de.caritas.cob.userservice.api.service;
+
+import de.caritas.cob.userservice.api.helper.AuthenticatedUser;
+import de.caritas.cob.userservice.api.model.AdminAgency;
+import de.caritas.cob.userservice.api.port.out.AdminAgencyRepository;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+import lombok.NonNull;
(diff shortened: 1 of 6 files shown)


Storypapst:
CodeRabbit review: **0 findings** across all 5 changed files (`AdminAuditAgencyScope`, both log services, and the three test classes).

shazia-k (approved):

PR #957, opened 2026-08-03
P0: finalize registered enquiry from encrypted Matrix event (#954)

Storypapst:
Fixes #954

## P0 problem

The registered-enquiry endpoint impersonated the Matrix user and posted plaintext `m.room.message` into an already encrypted room. Server impersonation has no browser crypto device and therefore bypassed Megolm.

## Fix

- Add `matrixEventId` to the enquiry-finalization contract.
- Remove server-side initial-message sending from `CreateEnquiryMessageFacade`.
- Read the raw event from the exact session room and accept it only when:
  - the event exists and its returned ID matches;
  - its type is `m.room.encrypted`;
  - its sender equals the authenticated asker's Matrix user ID.
- Only after validation update `message_date` / session status and emit email plus `request.new` notifications.
- Preserve appointment behavior.

The backend never receives the enquiry plaintext in the new flow.

## Verification

- `CreateEnquiryMessageFacadeTest`: 12/12
- `MatrixSynapseServiceTest`: 57/57
- `UserRegistrationControllerDelegateTest`: 24/24
- `AppointmentControllerTest`: 10/10
- `CreateEnquiryMessageFacadeMatrixRoomProvisioningTest`: 2/2
- Java 21, Spotless clean, `git diff --check` clean

The first full suite run executed 3,678 tests and found exactly two legacy provisioning assertions that still expected the removed plaintext server send. Those contracts were updated to validate the encrypted event instead and are green in the targeted rerun.

## Coordinated rollout

Deploy together with OpenResilienceInitiative/ORISO-Frontend#911. Backend first is contract-compatible with the new field, but old Frontend clients cannot finalize registered enquiries after this security hardening, so the deployment window must be coordinated.

Real-browser PreDev proof will be attached after the combined image is deployed.

## Summary by CodeRabbit

- **New Features**
  - Enquiry submissions now support validation of encrypted Matrix events supplied by the requester.
  - Added retrieval of specific Matrix room events for validation.
- **Bug Fixes**
  - Plaintext, missing, unauthorized, or invalid Matrix events are rejected.
  - Appointment enquiries continue to work without event validation.
- **Tests**
  - Added coverage for successful event retrieval and invalid or missing event scenarios.

PR #960, opened 2026-08-03
P0: stop double-decoding registration passwords (#959)

### LINKED ISSUE #959: [P0] Registration double-decodes passwords and creates unusable accounts
## Impact

Registration currently URL-decodes the password twice: once through UrlDecodePasswordJsonDeserializer on UserDTO.password and again in UserRegistrationControllerDelegate.registerUser.

Passwords containing percent-encoded-looking sequences can therefore be changed after registration. The API returns 201, but the user cannot log in with the submitted password (Keycloak invalid_grant). This blocked multiple valid Simpsons Test Access identities in PreDev.

## Expected

Decode the transport representation exactly once. The exact password produced by UserDTO deserialization must be passed unchanged to account creation and Keycloak.

## Reproduction

1. Register a user with a valid password containing a percent sequence that changes on a second URL decode.
2. Registration returns 201.
3. Log in using the original submitted password.
4. Keycloak rejects it with invalid_grant.

## Acceptance criteria

- Remove the second decode from the registration controller.
- Add a regression test proving a once-decoded password is passed unchanged.
- Keep URL decoding coverage at the JSON deserializer boundary.
- Run the UserService test suite.

Storypapst:
Closes #959.

## What changed

- remove the second URL decode in UserRegistrationControllerDelegate
- preserve the password value already produced by UserDTO deserialization
- update the controller regression test to prove percent sequences are not decoded a second time

## Why

UserDTO.password already uses UrlDecodePasswordJsonDeserializer. Decoding again in the controller can silently change legal passwords after registration, so registration returns 201 but subsequent Keycloak login fails.

## Verification

- red test first: controller changed pa%20ss to pa ss
- targeted controller and password-deserializer tests pass
- full Maven test suite passes
- Maven package passes
- git diff --check passes

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserRegistrationControllerDelegate.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserRegistrationControllerDelegate.java
@@ -29,8 +29,6 @@ import de.caritas.cob.userservice.api.service.auth.PasswordResetService;
 import de.caritas.cob.userservice.api.service.session.SessionService;
 import de.caritas.cob.userservice.api.service.user.UserAccountService;
 import jakarta.transaction.Transactional;
-import java.net.URLDecoder;
-import java.nio.charset.StandardCharsets;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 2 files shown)


Storypapst:
Live PreDev proof completed in run e2e-20260803-025044 with combined UserService commit 6d2279a5 (#957 plus this PR).

Jacqueline Bouvier was confirmed incompatible with the previous double-decode behavior. After deploying this fix:
- POST /service/users/askers/new returned 201
- immediate Keycloak token exchange returned 200
- a fresh, separate login with the same protected Test Access secret returned 200
- advice-seeker session 28 opened successfully
- live UserService pod Ready, 0 restarts

No password or recovery secret was logged or captured.

Shirloin (approved):

PR #981, opened 2026-08-07
fix(matrixrtc): resolve call policy in technical tenant context

Hassan9215:
## Problem

Element Call fails with **"Call access denied" (403)** for every user on predev, even when the user is a joined member of the session's Matrix room and all tenant call feature flags are enabled.

## Root cause

`/internal/matrixrtc/call-policy` is whitelisted from `HttpTenantFilter` (it is a cluster-internal, token-authenticated endpoint), so requests to it run **without tenant context**. `TenantAspect` fires before every repository call and, since the context is neither set nor technical, enables the Hibernate `tenantFilter` with `tenantId=null`:

```
(tenant_id = null OR (null = 1 AND tenant_id IS NULL))
```

This matches no rows, so `MatrixRtcPolicyContextResolver` never finds the session for the room, `MatrixRtcCallPolicyService` returns `CallMediaPolicy.denied()`, and the auth policy gateway fail-closes the call with 403 — silently: no log line on the whole path.

Verified on predev against session 78 / room `!WWoaRyNZzdCCHRRSXp`: the DB row has the correct `matrix_room_id` and `tenant_id=1`, tenant 1 has all call flags enabled, the user is one of 13 joined room members, yet the endpoint returned `{"audioAllowed":false,"videoAllowed":false}`.

## Fix

- Run the policy resolution under the technical tenant context (same `runCrossTenant` pattern as `AnonymousUsernameRegistry`), restoring the caller's tenant afterwards.
- Log the reason for every denial (non-member, unresolvable room, tenant lookup failure, flags off) so the next fail-closed 403 is diagnosable from userservice logs instead of requiring DB spelunking.

## Tests

- `resolvesRoomLookupInTechnicalTenantContext` — asserts the repository lookup observes `TECHNICAL_TENANT_ID` and the context is cleared afterwards (fails on pre-dev's current behavior).
- `restoresCallerTenantContextAfterResolution` — asserts a caller's tenant survives resolution.
- All 18 matrixrtc call-policy tests pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of cross-tenant call-policy checks by correctly managing tenant context.
  * Preserved the caller’s tenant context after policy evaluation completes.
  * Improved handling and diagnostics for tenant lookup failures and denied call-policy requests.
  * Added standardized, privacy-conscious denial reasons for call-policy decisions.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/service/matrixrtc/CallPolicyDenialReason.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/matrixrtc/CallPolicyDenialReason.java
@@ -8,6 +8,7 @@ package de.caritas.cob.userservice.api.service.matrixrtc;
  */
 public enum CallPolicyDenialReason {
   NOT_ROOM_MEMBER,
+  ROOM_MEMBERS_UNAVAILABLE,
   NO_TENANT_CONTEXT,
   TENANT_SETTINGS_UNAVAILABLE,
   CALLS_DISABLED_FOR_TENANT
--- a/src/main/java/de/caritas/cob/userservice/api/service/matrixrtc/CallPolicyDenialReason.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/matrixrtc/CallPolicyDenialReason.java
@@ -1,0 +1,14 @@
+package de.caritas.cob.userservice.api.service.matrixrtc;
+
+/**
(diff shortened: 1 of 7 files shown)


shazia-k (approved):
@Hassan9215 there was some data exposure printing in logs. I fixed that and committed

Issue #1010, opened 2026-08-14
[KDG] Privacy data lifecycle & retention


Storypapst:
## Why

The data-protection impact assessment (DPIA, August 2026) confirmed that counselling **content** is well protected: chat is end-to-end encrypted, notification previews are hard-disabled (`NONE` on every level), and admin statistics use small-cell suppression. What remains is a set of *data-household* gaps — plaintext metadata that outlives its purpose, tables without any retention, and rows that survive account deletion. None of these is an active leak of counselling content, with one exception (case-handover free-text, Task 1); all of them are findings a KDG/GDPR audit would raise. This epic closes them as one coherent package.

Design principle throughout: **not storing plaintext beats encrypting it** (data minimisation, Art. 5(1)(c) GDPR / § 7 KDG).

### 1. `event_notification`: finish the params-only migration (M, ~3–4 d)
The notification feed already renders from client-side i18n templates; the stored English `title`/`text` columns are a legacy bridge. One real content finding remains: the case-handover "explanation" free-text (written by counsellors, can reference case content) is persisted verbatim into `text`, indefinitely.
- [x] **1a** (S–M) Convert the 4 legacy producers to `params` and remove `explanation` from notifications entirely — it stays available on demand via the handover-request API (`caseHandoverRequestId` is already in `action_path`). Files: `api/service/CaseHandoverService.java` (`notifyGranted`, `notifyPendingConsent`, `notifyConsentDeclined`), `SendFinishedAnonymousConversationEventActionCommand.java`. *AC:* no `event_notification` row contains counsellor free-text; params carry `requesterName`, `reasonCode`, `sessionId`.
- [ ] **1b** (S) Frontend: pass parsed `params` as interpolation into `renderEventStrings` (`ORISO-Frontend/src/components/notificationsCenter/NotificationsCenter.tsx`) and verify the i18n keys per locale. *AC:* `{{senderDisplayName}}`-style templates render with real values.
- [ ] **1c** (S) Server stops writing/serving `title`/`text` (deploy after 1b). File: `api/service/notification/EventNotificationService.java` (`buildEvent`, `toItem`). *AC:* new rows have empty display-text columns; feed renders correctly from params alone.
- [x] **1d** (S) Backfill: one-off `UPDATE event_notification SET title='', text=NULL` (Liquibase; Pre-Dev may need manual execution); column drop as a later follow-up. *AC:* no historical plaintext remains.
[...]

PR #1015, opened 2026-08-14
Age event notifications out instead of keeping them forever

Storypapst:
## Why

`event_notification` had no retention at all. Every row carries the recipient's Keycloak UUID, a session reference and third-party names, and `read_date` is a second-precision record of when the recipient looked at a given notification — an activity profile that grew for as long as the account existed. None of that has a purpose once the notification has been seen.

## What changes

A nightly purge with **two cutoffs**, not one. A read notification has already done its job, so it goes after the shorter period; the absolute cutoff makes sure an unread feed cannot grow without bound either.

| Setting | Default | Environment variable |
| --- | --- | --- |
| `event.notification.retention.read.days` | 90 | `EVENT_NOTIFICATION_RETENTION_READ_DAYS` |
| `event.notification.retention.absolute.days` | 365 | `EVENT_NOTIFICATION_RETENTION_ABSOLUTE_DAYS` |

The defaults are the DPIA proposal. They are configuration rather than constants because the periods still await sign-off by the client's data-protection officer — this PR ships the mechanism, and the numbers can change without a code change.

## Two decisions worth reviewing

**A non-positive period disables its cutoff — it does not purge everything.** Read literally as "0 days of retention", the obvious implementation wipes the table on an empty or mistyped value. That reading is closed off explicitly and pinned by tests, because the failure mode is silent and total.

**Bulk JPQL rather than derived deletes.** A derived `deleteBy…` loads every expired row into the persistence context and removes them one at a time; the first run after deployment could touch a lot of rows.

The scheduler follows `DeleteUserAccountScheduler`: every replica runs the same cron, so the run is claimed through `ScheduledTaskClaimService` and exactly one replica does the deleting.

## Verification

- `EventNotificationRetentionRepositoryIT` executes both queries against the H2 testing schema. Nothing short of a real execution proves a bulk delete selects the rows it is meant to, and getting the predicate wrong here deletes a user's whole feed. It also pins both cutoffs as **exclusive**, so a row exactly on the boundary is still inside its retention period.
- `EventNotificationRetentionServiceTest` covers both cutoffs, their relative ordering, and each disable path independently.
- `EventNotificationRetentionSchedulerTest` covers the claim being won and lost.

Part of #1010 — task 2a.

## Summary by CodeRabbit

- **New Features**
  - Added automatic cleanup of expired event notifications.
  - Read notifications older than 90 days are removed, while all notifications older than 365 days are purged by default.
  - Cleanup runs daily and can be configured or disabled.

- **Reliability**
  - Coordinated execution prevents duplicate cleanup runs across service instances.

- **Tests**
  - Added coverage for retention periods, cutoff boundaries, unread notifications, disabled settings, and scheduled execution.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/port/out/EventNotificationRepository.java
+++ b/src/main/java/de/caritas/cob/userservice/api/port/out/EventNotificationRepository.java
@@ -1,10 +1,14 @@
 package de.caritas.cob.userservice.api.port.out;
 
 import de.caritas.cob.userservice.api.model.EventNotification;
+import java.time.LocalDateTime;
 import java.util.List;
 import java.util.Optional;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.data.jpa.repository.Modifying;
+import org.springframework.data.jpa.repository.Query;
+import org.springframework.data.repository.query.Param;
 
 public interface EventNotificationRepository extends JpaRepository<EventNotification, Long> {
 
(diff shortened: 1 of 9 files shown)


Storypapst:
@coderabbitai review

PR #802, opened 2026-07-27
[Stability] Focus identity account settings

Storypapst:
## Summary

- introduce the provider-neutral `IdentityAccountSettingsUpdater` output port
- route current-account password and preferred-language changes through the focused port
- remove both operations from the broad `IdentityClient`, leaving general provisioning/password-reset writes unchanged
- remove the last broad-client dependency from `IdentityManager`
- protect the boundary and shared Spring identity mocks with executable contracts

## Measured call bounds

- current-account password attempt: one user-resource resolution and at most one password reset; existing boolean success/failure behavior is preserved
- preferred-language change: one representation read and at most one update
- unchanged preferred language: one representation read and no update
- general provisioning and password-reset consumers retain the existing `updatePassword` command

## Verification

- `./mvnw -B -Dskip.integration-tests=true clean test`
  - 3,868 tests, 0 failures, 0 errors, 0 skipped
- `ORISO_LOCAL_REDIS_IT=true ./mvnw -B -Dskip.unit-tests=true clean integration-test`
  - 969 tests, 0 failures, 0 errors, 5 conditional infrastructure skips
- focused manager, Keycloak and web tests
  - 154 tests, 0 failures, 0 errors, 0 skipped
- `python3 -m unittest discover -s tests/ci -p 'test_*.py'`
  - 42 tests, all passing
- `python3 scripts/ci/check-no-test-quarantine.py`
  - no quarantine annotations
- `./mvnw -B spotless:check`
- `git diff --check`

## Scope and rollout

This is a modular-monolith boundary change stacked on #800. It does not extract
a microservice, merge, or deploy anything. PreDev runtime and SigNoz evidence
remain separate release gates after the stack is reviewed and merged.

The product target remains Matrix-only chat with the ORISO frontend, a
controlled MatrixRTC/Element Call fork, and LiveKit. Rocket.Chat and Jitsi are
legacy removal targets, not fallback architecture.

Closes #801

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #801, opened 2026-07-27
Complete the focused identity account-settings boundary


Storypapst:
Parent: #335

## Why

The remaining current-account password and preferred-language mutations still cross the broad `IdentityClient`. Both active application calls are mediated exclusively by `IdentityManager`, while the broad client also owns unrelated provisioning, role-write and lifecycle commands. The two account-setting operations already have distinct, measurable provider call bounds and can be separated without moving the general provisioning/password-reset command.

## What to build

- Introduce a provider-neutral account-settings output port for current-account password changes and preferred-language updates.
- Route both `IdentityManager` operations through the focused port.
- Remove `changePassword` and `changeLanguage` from the broad `IdentityClient`.
- Preserve `updatePassword` on the provisioning/reset boundary for its existing consumers.
- Keep Keycloak representation handling, password-policy behavior and update/no-op decisions inside the adapter.
- Add executable boundary, interaction and shared Spring-context contracts.
- Update the stability documentation and exact call bounds.

## Acceptance

- [x] `IdentityManager` no longer depends on the broad `IdentityClient`.
- [x] A current-account password attempt performs exactly one user-resource resolution and at most one password reset, preserving the existing boolean success/failure behavior.
- [x] A preferred-language update performs exactly one representation read and at most one update.
- [x] An unchanged preferred language performs no update after the single read.
- [x] `IdentityClient` exposes neither current-account password change nor preferred-language mutation.
- [x] General provisioning/password-reset consumers retain their existing `updatePassword` behavior.
- [x] Complete Java 21 unit/integration and CI-contract suites remain green with zero quarantine.
- [x] Stability documentation records the focused boundary and exact call bounds.

## Implementation

Draft PR #802, commit `9fa2db01`.

- `IdentityManager` now consumes the focused `IdentityAccountSettingsUpdater` and has no broad-client dependency.
- Local: 3,868 unit tests and 969 integration/contract/E2E tests are green.
- Focused manager, Keycloak and web tests: 154 green.
- CI contracts: 42 green; zero quarantine; Spotless and diff checks green.
- GitHub: 11/11 checks green, including MariaDB, Redis, full integration, Docker validation, API diff and required PreDev CI.
- Normal local MariaDB was not reset; the dedicated local Redis test container remained healthy.

[...]

PR #800, opened 2026-07-27
[Stability] Remove unused identity session-close surface

Storypapst:
## Summary

- remove the unused `closeSession` command from the broad `IdentityClient`
- remove the dead forwarding methods from `KeycloakService` and `KeycloakAuthClient`
- remove the two tests that only exercised the unused forwarding chain
- add an executable boundary contract that prevents the dead command surface from returning
- update the stability inventory and exact call bound

## Call-bound evidence

Repository-wide production tracing found no application consumer of
`closeSession`. The removed path therefore had an exact zero-call bound. The
active refresh-token logout flow remains unchanged and retains its focused
`IdentityAuthentication` boundary.

## Verification

- `./mvnw -B -Dskip.integration-tests=true clean test`
  - 3,866 tests, 0 failures, 0 errors, 0 skipped
- `ORISO_LOCAL_REDIS_IT=true ./mvnw -B -Dskip.unit-tests=true clean integration-test`
  - 969 tests, 0 failures, 0 errors, 5 conditional infrastructure skips
- focused Keycloak tests
  - 110 tests, 0 failures, 0 errors, 0 skipped
- `python3 -m unittest discover -s tests/ci -p 'test_*.py'`
  - 41 tests, all passing
- `python3 scripts/ci/check-no-test-quarantine.py`
  - no quarantine annotations
- `./mvnw -B spotless:check`
- `git diff --check`

## Scope and rollout

This is dead-surface removal inside the modular monolith, stacked on #798. It
does not change active identity behavior, extract a microservice, merge, or
deploy anything. PreDev runtime and SigNoz evidence remain separate release
gates after the stack is reviewed and merged.

The product target remains Matrix-only chat with the ORISO frontend, a
controlled MatrixRTC/Element Call fork, and LiveKit. Rocket.Chat and Jitsi are
legacy removal targets, not fallback architecture.

Closes #799

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #796, opened 2026-07-27
[Stability] Complete identity role membership read boundary

Storypapst:
## Summary

- route `IdentityManager.hasRole` through the existing focused `IdentityRoleLookup`
- perform exactly one external full-role list read and compare the requested application role in-process
- remove `userHasRole` and the unused `userHasAuthority` evaluator from the broad `IdentityClient` boundary
- remove three obsolete authority tests and stale role-check test stubs
- update the shared Keycloak integration-test double for the focused role lookup
- strengthen the executable module-boundary contract and stability documentation

## Verification

- JDK 21 unit suite: 3,867 tests, 0 failures, 0 errors, 0 skipped
- integration + contract + E2E suite: 969 tests, 0 failures, 0 errors, 5 environment-gated skips
- focused Java tests: 125 tests, 0 failures, 0 errors, 0 skipped
- affected Spring integration tests: 12 tests, 0 failures, 0 errors, 0 skipped
- isolated regression test after correcting the shared test double: 1 test, green
- Python structural contracts: 39 tests, green
- zero-quarantine checker: green
- Spotless and `git diff --check`: green

## Scope

- stacked on draft PR #794
- continues the modular-monolith-first boundary work; it does not extract a service
- nothing is merged or deployed by this PR
- live PreDev and SigNoz call-frequency/latency evidence remains a later release gate
- the target architecture is Matrix chat through the ORISO frontend with an ORISO-controlled MatrixRTC/Element Call fork and LiveKit; Rocket.Chat and Jitsi are removal targets, not fallback dependencies

Closes #795

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1039, opened 2026-08-16
[Stability] Focus identity account settings (replay of #802)

shazia-k:
Replay of #802 directly onto `pre-dev`, on top of #1038 (#796's replay). The original targets `refactor/remove-identity-session-close-799`, whose PR (#800) was closed.

## Change

`changePassword` and `changeLanguage` sat on the broad `IdentityClient`, next to user creation and deletion — a caller that only wanted to change a language held the whole provisioning surface. They move to a focused `IdentityAccountSettingsUpdater`, and with them the **last broad-client dependency leaves `IdentityManager`**.

`changeLanguage` is renamed `changePreferredLanguage`; the adapter tests and the shared Spring double follow.

## Checked before applying

`IdentityManager` is the only caller of either method in `main`, so removing them from the broad port strands nothing.

This PR depends on #1038 — it deletes the `IdentityClient` field outright, which only compiles once `hasRole` no longer uses it.

## Test wiring

Follows this repo's existing pattern rather than adding beans: the identity mock in each Spring test declares its ports via `extraInterfaces`, so `IdentityAccountSettingsUpdater.class` joins that list in the seven affected tests. (I first added a separate `@MockitoBean`, then replaced it after reading how the original PR and the surrounding tests do it.)

`UserControllerIT`'s "password was never changed" assertion is **kept**, not dropped — it now verifies against the focused port, since that is where the self-service change lives.

## Ratchets

Two, both confirmed to bite by reintroducing the regression and watching them fail:
- `IdentityClient` may not carry `changePassword` / `changeLanguage`
- `IdentityManager` may not import the broad client at all

## Verification

- `./mvnw clean test` → **3,954 unit tests, 0 failures, 0 errors**
- `suite-inventory.py --require unit` → 3,954 across 450 reports
- CI contracts → **99 tests, OK**
- Spotless clean

Refs #801, parent #335. Supersedes #802.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/IdentityManager.java
+++ b/src/main/java/de/caritas/cob/userservice/api/IdentityManager.java
@@ -6,8 +6,8 @@ import de.caritas.cob.userservice.api.identity.IdentityEmailVerification;
 import de.caritas.cob.userservice.api.identity.IdentityEmailVerificationStart;
 import de.caritas.cob.userservice.api.identity.IdentityOtpCredential;
 import de.caritas.cob.userservice.api.port.in.IdentityManaging;
+import de.caritas.cob.userservice.api.port.out.IdentityAccountSettingsUpdater;
 import de.caritas.cob.userservice.api.port.out.IdentityAuthentication;
-import de.caritas.cob.userservice.api.port.out.IdentityClient;
 import de.caritas.cob.userservice.api.port.out.IdentityEmailAddressUpdater;
 import de.caritas.cob.userservice.api.port.out.IdentityEmailOwnerLookup;
 import de.caritas.cob.userservice.api.port.out.IdentityRoleLookup;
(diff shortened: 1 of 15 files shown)


kiodreambau:
Parked for now, by the project owner's decision on 2026-08-17: the `replay/identity-*` lane is frozen until ORISO-UserService#888 lands.

Reason: #888 and this lane are doing the same identity-port work in parallel. #888 sat `CONFLICTING` for 20 days, and its conflicts are exactly on the files the merged replays rewrote — `IdentityManager.java`, `KeycloakService.java`, `IdentityClient.java`. Running both lanes is what produced the conflict in the first place, so they cannot converge while both are active.

#888 is now `MERGEABLE` again with the full suite green (`Tests run: 4074, Failures: 0`). Once it merges, the identity-port work it carries makes this PR unnecessary; if anything is left over afterwards, it is smaller than what is here today.

Nothing in this PR was rejected on its merits — this is sequencing only.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Issue #1060, opened 2026-08-18
Late-joiner counsellors stay locked out of existing enquiries after joining an agency


kiodreambau:
**Affected repos:** ORISO-UserService

## Why

OpenResilienceInitiative/ORISO-UserService#905 fixed the enquiry visibility bug by joining all agency counsellors into the enquiry Matrix room **at enquiry creation**. The gap that remains (reported in OpenResilienceInitiative/ORISO-Frontend#811, requested as its own urgent ticket by @BjoernLudwig on 2026-08-13): a counsellor who is added to an agency **after** an enquiry already arrived is never joined to the existing enquiry rooms, so they stay locked out of open initial requests.

## What to build

- When a counsellor is assigned to an agency, join them into the Matrix rooms of that agency's existing open enquiries (the "late-joiner hook"), mirroring the membership logic of OpenResilienceInitiative/ORISO-UserService#905.
- Cover the reverse path where relevant: removal from the agency must not leave stale memberships behind.
- Regression test: counsellor added after enquiry creation can read the enquiry (text + attachments) exactly like a counsellor who was present at creation.

## Acceptance

- [x] Counsellor added to an agency after an enquiry exists can open and read that enquiry without manual intervention.
- [x] Existing at-creation behavior from OpenResilienceInitiative/ORISO-UserService#905 unchanged.
- [x] Integration test covers the late-joiner path and fails before the fix.

## How should I test this? (reviewer plan)

- [ ] Create an enquiry for agency A → add a NEW counsellor to agency A → log in as that counsellor → the enquiry is visible and readable (incl. attachments).
- [ ] Counsellor present at creation still sees the enquiry unchanged.

kiodreambau:
## Status: fix is on `pre-dev` in OpenResilienceInitiative/ORISO-UserService#1074

Worth recording why this issue was still open: OpenResilienceInitiative/ORISO-UserService#1063 already implemented it and is merged — but into `speed/reflux-progress-exp-16-aug`, not `pre-dev`. That branch is stuck behind a red openapi consumer check, so the code never reached `pre-dev` and the bug stayed live there. #1074 brings it onto `pre-dev` and closes four gaps #1063 left open.

### Acceptance criteria

- [x] Counsellor added to an agency after an enquiry exists can open that enquiry without manual intervention — `AgencyLateJoinerMembershipService`, hooked into both the atomic and the legacy two-step assignment path.
[...]

PR #905, opened 2026-07-28
feat(matrix): join the agency's counsellors into the enquiry room at creation (FE#811)

Storypapst:
Refs OpenResilienceInitiative/ORISO-Frontend#811 (P0). Implements ADR-002 §1 "real members from room creation". Companion FE PR: retry failed decryptions after a history-key import (`fix/enquiry-history-key-retry-811`) — **merge both**; each alone is insufficient.

## Problem

A counsellor reviewing an initial request saw an **empty conversation** — no messages, no files, no error. Root cause: the enquiry holding room was created with only the agency service account and the advice seeker as members; counsellors were invited at **accept** time. A non-member's Matrix `/sync` never contains the room, so the client had literally nothing to render.

## Change

- New `AgencySilentMembershipService`: invites + joins every active counsellor of the enquiry's agency into the holding room **at room creation, before the first message can be sent**. Ordering is load-bearing: a Megolm session only reaches devices present at encrypt time; anyone joined later cannot decrypt earlier content.
- Called from `AgencyPreAssignmentRoomService.ensureHoldingRoom` (both the registration and the enquiry-message path).
- Directly addressed enquiries (`isConsultantDirectlySet`) are excluded — the advice seeker chose one counsellor; the case must not fan out to the department.
- Best-effort per counsellor: an unresolvable Matrix account never costs the other counsellors their membership and never fails the advice seeker's enquiry.
- Provisions missing counsellor Matrix accounts (same create-or-resolve probe as `AssignEnquiryFacade`).

## Tests

- `AgencySilentMembershipServiceTest` (7 new): full-join, account provisioning, login-probe resolution, invite-rejection tolerance, per-consultant best effort, skip-on-unresolvable, input short-circuit.
- `AgencyPreAssignmentRoomServiceTest` (+2): join happens between room creation and persist; direct enquiries skip the fan-out.
- `CreateEnquiryMessageFacadeMatrixRoomProvisioningTest` (+1): counsellor is a member **before** the enquiry message is sent (in-order verification).
- Full unit suite: **3437 tests, 0 failures**.

## Live proof (Pre-Dev, run `_e2e-artifacts/e2e-20260728-fe811`)

| Check | Before | After (session 103532, with FE companion) |
|---|---|---|
| Room members at enquiry creation | 3 (agency svc, asker, matrixadm) | 8 (+ all 5 counsellors of agency 241) |
| Counsellor opens enquiry before accept | empty pane | full plaintext content |
[...]

PR #1055, opened 2026-08-17
Fail required integration CI on skipped tests

Storypapst:
Closes OpenResilienceInitiative/ORISO-UserService#1054

## Summary

- count `executed = tests - skipped` against the existing 830-test floor
- fail the required integration contract on every skipped report while retaining report, failure, error, and critical-E2E guards
- activate the Redis-gated contract in every full required job
- keep real-MariaDB contracts in the separately required MariaDB job and exclude them from primary report discovery, including the deliberately out-of-refactor-CI support-room convergence test from commit `3bfa3d06`

## Verification

Local only, on macOS with Temurin JDK 21.0.11, Redis 7, and MariaDB 10.11.

- RED: synthetic 830-total/4-skipped reports returned exit 0 before the guard change
- GREEN: 99 CI contract tests passed
- GREEN: required integration runner reported `reports=86 tests=882 executed=882 failures=0 errors=0 skipped=0`
- GREEN: authoritative MariaDB contract reported `tests=12 failures=0 errors=0 skipped=0`
- GREEN: workflow YAML parsed, shell syntax passed, and `git diff --check` passed

### Reviewer test plan

- [ ] Run the required integration workflow and confirm it reports executed tests separately with zero skips.
- [ ] Feed a synthetic skipped report to the guard and confirm it exits non-zero for both the executed floor and skipped count.
- [ ] Remove a critical E2E report from the fixture and confirm the existing critical-report check still fails.
- [ ] Run the MariaDB and Redis contracts and confirm both required infrastructure paths remain green.

Review focus: `shazia-k` for the CI contract and UserService suite ownership; `Shirloin` for workflow aggregation and the infrastructure split.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Repaired support-room database structures when missing, improving migration reliability.
  * Removed draft records that contain no meaningful text after markup cleanup.
* **Tests**
  * Improved Redis-backed integration-test validation across development, pull request, and main-branch checks.
  * Expanded MariaDB contract coverage to include migration convergence scenarios.
  * Refined test reporting requirements to verify sufficient coverage and fail on errors, failures, or skipped tests.
* **Documentation**
  * Updated the integration-test inventory and clarified how application, Redis, and MariaDB checks are evaluated.

PR #1063, opened 2026-08-18
Join late-joining counsellors into existing enquiry rooms (and revoke on removal)

kiodreambau:
## What & why

OpenResilienceInitiative/ORISO-UserService#905 joins the agency's counsellors into an enquiry's Matrix room **at room creation**. That only helps counsellors who were already in the agency at that moment. A counsellor added to the agency **afterwards** never becomes a member of the rooms that already exist — Matrix `/sync` returns nothing for a non-member, and no later event repairs it, so open initial requests stay invisible to them until an admin intervenes.

This PR adds `AgencyLateJoinerMembershipService` and wires it into both directions of the agency relation:

- **Join (the reported bug):** completing an agency assignment backfills the counsellor into the Matrix rooms of the agency's existing open enquiries. Both entry points are covered — the atomic `createNewConsultantAgency` path the admin UI uses and the legacy two-step `completeConsultantAgencyAssigment` path — because they funnel through one place.
- **Removal (the reverse path the issue asks for):** marking a consultant-agency relation deleted revokes that membership again, so a counsellor does not keep syncing enquiries of an agency they left. The hook sits in `markAsDeleted`, which covers both the single-agency endpoint and the bulk path that `ConsultantAdminFacade#setConsultantAgencies` uses when an admin edits a consultant's agency set.

**Scope:** the agency's *open* enquiries — status `NEW`, no counsellor assigned, not directly addressed. Those are the cases the whole department is entitled to pick up; an accepted case belongs to one counsellor and is governed by `AssignEnquiryFacade`. On the removal side that scope is not a preference but a constraint: `AssignEnquiryFacade` removes the agency service account from the room when an enquiry is accepted, so the agency token can no longer act on an accepted case at all.

**Failure semantics:** both directions are best effort. The relation in the database is the source of truth, so a Synapse failure degrades to "membership repaired later" rather than failing the admin's request. Per-counsellor membership logic is shared with the at-creation fan-out by promoting `AgencySilentMembershipService#joinSilently` to the public `joinConsultantIntoRoom`, so the backfill cannot drift from #905's semantics.

Closes OpenResilienceInitiative/ORISO-UserService#1060
Refs OpenResilienceInitiative/ORISO-Frontend#811

## How should I test this?

- [ ] Create an enquiry for agency A (as an advice seeker) → confirm a counsellor already in agency A sees it → **expected:** unchanged behaviour from #905, enquiry visible and readable.
- [ ] Add a **new** counsellor to agency A in the admin panel → log in as that counsellor → open the initial-request list → **expected:** the enquiry created *before* they joined is listed, opens, and its message text is readable.
- [ ] Same enquiry, check attachments → **expected:** attachments load for the late joiner.
- [ ] Check the service log during the assignment → **expected:** one `Backfilled consultant <id> into N of N open enquiry rooms of agency <id>` line.
- [ ] Remove that counsellor from agency A again → log in as them → **expected:** the enquiry disappears from their list; log shows `Removed consultant <id> from N of N open enquiry rooms of agency <id>`.
- [ ] Edit a consultant in the admin panel and drop one agency from their agency set (the bulk path) → **expected:** the same revocation happens for the dropped agency, and agencies left untouched keep their enquiries.
- [ ] Assign an agency whose Matrix service account is misconfigured (or stop Synapse) → **expected:** the assignment request still succeeds with 2xx; only a `WARN` is logged. Nothing 500s.
- [ ] Create an enquiry via a **direct** counsellor link, then add a new counsellor to the agency → **expected:** the directly addressed enquiry is *not* fanned out to them.

## Verified locally vs CI must prove

**Verified locally** (JDK 21 via `brew --prefix openjdk@21`, `./mvnw`, targeted test classes only):

| Test class | Result |
| --- | --- |
| `AgencyLateJoinerMembershipServiceTest` (new) | `Tests run: 9, Failures: 0, Errors: 0, Skipped: 0` |
| `ConsultantAgencyRelationCreatorServiceTest` | `Tests run: 16, Failures: 0, Errors: 0, Skipped: 0` |
| `ConsultantAgencyAdminServiceTest` | `Tests run: 14, Failures: 0, Errors: 0, Skipped: 0` |
| `ConsultantAgencyAdminUserServiceTest` | `Tests run: 7, Failures: 0, Errors: 0, Skipped: 0` |
| `AgencySilentMembershipServiceTest` (#905, untouched behaviour) | `Tests run: 7, Failures: 0, Errors: 0, Skipped: 0` |
| `AgencyPreAssignmentRoomServiceTest` (#905, untouched behaviour) | `Tests run: 16, Failures: 0, Errors: 0, Skipped: 0` |
| Aggregate | `Tests run: 69, Failures: 0, Errors: 0, Skipped: 0` — `BUILD SUCCESS` |

`mvn spotless:apply` is clean (google-java-format, GOOGLE style).

**Failing-test proof.** The tests were confirmed to fail without the fix, not merely to pass with it. Both call sites were commented out (`backfillExistingEnquiryRooms` in `ConsultantAgencyRelationCreatorService`, `revokeOpenEnquiryRoomMembership` in `ConsultantAgencyAdminService`) and both public methods of `AgencyLateJoinerMembershipService` were made to return `0` immediately. Result: `Tests run: 39, Failures: 9, Errors: 3` — `BUILD FAILURE`, including

```
ConsultantAgencyRelationCreatorServiceTest.createNewConsultantAgency_Should_backfillTheConsultantIntoExistingOpenEnquiryRooms
ConsultantAgencyRelationCreatorServiceTest.completeConsultantAgencyAssigment_Should_backfillTheConsultantIntoExistingOpenEnquiryRooms
ConsultantAgencyAdminServiceTest.markConsultantAgencyForDeletion_Should_RevokeOpenEnquiryRoomMembership
ConsultantAgencyAdminServiceTest.markConsultantAgenciesForDeletion_Should_RevokeMembershipForEveryRemovedAgency
AgencyLateJoinerMembershipServiceTest.joinConsultantIntoOpenEnquiryRooms_joinsEveryOpenEnquiryRoomOfTheAgency:96 expected: <2> but was: <0>
AgencyLateJoinerMembershipServiceTest.removeConsultantFromOpenEnquiryRooms_removesFromEveryOpenEnquiryRoom:187 expected: <2> but was: <0>
```

The disabling edits were reverted from backups and the suite was re-run green (`Tests run: 69, Failures: 0` — `BUILD SUCCESS`) before committing.

**CI must prove** (deliberately not run locally — 16 GB machine, targeted classes only):

- The **full unit suite** — nothing outside the six classes above was executed here. Every Spring component that constructs `ConsultantAgencyAdminService` or `ConsultantAgencyRelationCreatorService` gained a constructor argument, so any test building them by hand rather than via `@InjectMocks` will surface in CI, not locally.
- **Spring context startup** — `AgencyLateJoinerMembershipService` is a new `@Service` with four injected collaborators; only a full context load proves the wiring.
- **Integration tests (`*IT`)** — none were run.
- **A real Matrix round-trip.** Every assertion here is against mocks. That invite → login → join actually produces a readable room for the late joiner, and that `removeUserFromRoom` with the *agency* token actually kicks a counsellor from an unassigned enquiry room, is only proven on pre-dev by the reviewer checklist above.

## Open risks and scope explicitly not covered

1. **Rollback amplifier.** `ConsultantAdminFacade#setConsultantAgencies` is `@Transactional` and performs the bulk removal and the creations inside one transaction. Matrix membership changes are not transactional: if that transaction rolls back later (e.g. the subsequent topic validation fails), the DB relations are restored but the Matrix joins/kicks are not undone. This PR keeps the side effects inline because that is what every existing Matrix side effect in this codebase does and there is no after-commit hook idiom here to follow; introducing one is a separate change. Worth a follow-up ticket.
2. **Stale membership in already-accepted rooms is not cleaned up.** A counsellor who was a silent member at creation, whose colleague then accepted the case, keeps that membership after leaving the agency. This is deliberately out of scope: once the enquiry is accepted the agency service account has left the room, so the agency token cannot remove anyone. Fixing it needs the assigned counsellor's or an admin token — a different mechanism (see the case-handover / team-session paths).
3. **Latency.** The backfill runs synchronously inside the admin's assignment request and costs roughly one invite + one login + one join per open enquiry room. An agency with a large backlog of open enquiries will make that request noticeably slower. Acceptable for correctness first; a queue is the follow-up if it bites.
4. **Megolm history.** A counsellor backfilled after messages were already sent holds no Megolm key for them and depends on the client-side history-key transfer shipping in OpenResilienceInitiative/ORISO-Frontend#811. Joining at creation (#905) remains the stronger guarantee — this is the repair path for everyone who could not be there at creation. Reviewers should read old messages, not just confirm the room appears.
5. **Base branch.** This targets `speed/reflux-progress-exp-16-aug` (the shipping branch) on purpose, not `pre-dev`.

### DIFF
--- a/documentation/user-service-historical-failure-classification.json
+++ b/documentation/user-service-historical-failure-classification.json
@@ -196,19 +196,19 @@
       "primaryTotal": -509
     },
     "verifiedCurrent": {
-      "verifiedApplicationHead": "374642585ffaa8ebca1d20c1631f0a9373eacf6a",
-      "unit": 4107,
+      "verifiedApplicationHead": "e6af9a1bf6540a7c7d3a54318bc679ca9e457a0a",
+      "unit": 4122,
       "integration": 913,
-      "primaryTotal": 5020,
+      "primaryTotal": 5035,
       "localEvidence": [
         "JDK 21: ./mvnw -Dskip.integration-tests=true test plus exact inventory gate",
         "JDK 21: scripts/ci/run-required-integration-tests.sh plus exact inventory gate"
       ]
     },
     "sinceCutoverDelta": {
-      "unit": 734,
+      "unit": 749,
       "integration": 73,
-      "primaryTotal": 807
+      "primaryTotal": 822
     },
     "sourceDiff": {
       "fromCommit": "dce9b2f4d89851d452135de6b0697d22413368c7",
@@ -221,8 +221,8 @@
   },
   "currentRequiredSuite": {
     "command": "scripts/ci/run-required-integration-tests.sh",
-    "unitTests": 4107,
-    "unitReports": 457,
+    "unitTests": 4122,
+    "unitReports": 458,
     "unitSkipped": 0,
     "integrationReports": 99,
     "integrationTests": 913,
--- a/src/main/java/de/caritas/cob/userservice/api/admin/service/agency/ConsultantAgencyAdminService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/admin/service/agency/ConsultantAgencyAdminService.java
@@ -22,18 +22,21 @@ import de.caritas.cob.userservice.api.port.out.ConsultantAgencyRepository;
 import de.caritas.cob.userservice.api.port.out.ConsultantRepository;
 import de.caritas.cob.userservice.api.port.out.SessionRepository;
 import de.caritas.cob.userservice.api.service.agency.AgencyService;
+import de.caritas.cob.userservice.api.service.session.AgencyLateJoinerMembershipService;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.beanutils.BeanUtils;
 import org.springframework.stereotype.Service;
 
 /** Service class to handle administrative operations on consultant-agencies. */
 @Service
 @RequiredArgsConstructor
+@Slf4j
 public class ConsultantAgencyAdminService {
 
   private final @NonNull ConsultantAgencyRepository consultantAgencyRepository;
@@ -43,6 +46,7 @@ public class ConsultantAgencyAdminService {
   private final @NonNull AgencyService agencyService;
   private final @NonNull AgencyAdminService agencyAdminService;
   private final @NonNull ConsultantAgencyDeletionValidationService agencyDeletionValidationService;
+  private final @NonNull AgencyLateJoinerMembershipService agencyLateJoinerMembershipService;
 
   /**
    * Returns all Agencies for the given consultantId.
@@ -250,6 +254,33 @@ public class ConsultantAgencyAdminService {
     this.agencyDeletionValidationService.validateAndMarkForDeletion(consultantAgency);
     consultantAgency.setDeleteDate(nowInUtc());
     this.consultantAgencyRepository.save(consultantAgency);
+    revokeOpenEnquiryRoomMembership(consultantAgency);
+  }
+
+  /**
+   * US#1060, symmetric half: since a counsellor is joined into the agency's open enquiry rooms when
+   * they are assigned, detaching them again has to take that membership back. Otherwise the
+   * counsellor keeps receiving the agency's open initial requests through Matrix {@code /sync} long
+   * after the relation was deleted.
+   *
+   * <p>Best effort — the deleted relation is already persisted and is the source of truth; a
+   * Synapse failure must not turn a successful detach into an error response.
+   */
+  private void revokeOpenEnquiryRoomMembership(ConsultantAgency consultantAgency) {
+    var consultant = consultantAgency.getConsultant();
+    if (consultant == null) {
+      return;
+    }
+    try {
+      agencyLateJoinerMembershipService.removeConsultantFromOpenEnquiryRooms(
+          consultant, consultantAgency.getAgencyId());
+    } catch (RuntimeException ex) {
+      log.warn(
+          "Could not revoke enquiry room membership of consultant {} in agency {}: {}",
+          consultant.getId(),
+          consultantAgency.getAgencyId(),
+          ex.getMessage());
+    }
   }
 
   /**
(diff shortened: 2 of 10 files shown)


kiodreambau:
### Note for reviewers: the red `openapi provider and consumer contracts` check is inherited, not caused by this PR

This branch changes **only Java files** (8 files, no OpenAPI YAML — see the commit stat). The failing step is `Verify TenantService admin consumer compatibility`, which reports:

```
in API POST /tenant/public/dpa/confirm/{token} removed the optional property `dpaVersion` from the response with the `200` status
in API GET /tenantadmin/{id}/dpa/signatures removed the optional property `items/dpaVersion` from the response with the `200` status
```

That mismatch comes from the base branch `speed/reflux-progress-exp-16-aug`, whose `services/tenantadminservice.yaml` expects `dpaVersion` while the pinned TenantService provider spec no longer returns it. The consumer contract was brought in by an earlier merge on the base branch (`913b7003`), not by this work.

Evidence:
- Open PRs based on `pre-dev` (#1059, #1055, #1045, #1041) pass this check; this PR is the only one based on the shipping branch.
- The contract workflow only runs on pull requests, so the base branch itself never surfaced the failure on push.

Nothing in this PR can fix it — the base branch needs either a refreshed provider pin or a corrected `services/tenantadminservice.yaml`. Worth its own ticket before the shipping branch is promoted.

kiodreambau:
### CI is green except the inherited contract check

Final run on `518ff5e3`: **11 checks pass, 1 fails.**

Passing: `test, build and Docker validation`, `test and build UserService`, `required integration tests` (both lanes), `mariadb-contract` (both lanes), `Redis shared-state contracts`, `openapi contract gate tests`, `api-contract-diff`, `required PreDev CI`, CodeRabbit.

Failing: `openapi provider and consumer contracts` — inherited from the base branch, not caused by this PR. Now tracked as OpenResilienceInitiative/ORISO-UserService#1064 with the full evidence; it cannot be fixed from here.

The earlier failure of the two build lanes was the exact-inventory gate (`expected 4107 tests, found 4122`): all 4,122 unit tests passed, only the recorded baseline was stale. Commit `518ff5e3` records the new inventory (4,122 unit tests across 458 reports) following the same pattern as `bc51c8e9`.

shazia-k:
The Error in this PR is not a blocker. and this PR is mergaeable.
reason is here
the pin at openapi-contracts.yml:56 points at a TenantService commit that is 2 ahead / 34 behind pre-dev and predates 164d130d, which is where dpaVersion actually landed. Nothing in PR 1063 is involved; it changes no YAML and no workflow.

PR #1066, opened 2026-08-19
feat(account-invite): refuse invite creation for an already registered e-mail

kiodreambau:
Owner-reported problem 3 of four, **backend half**. UI counterpart: ORISO-Admin `feat/invite-email-duplicate-check-p3`. See `0 - Docs/plans/2026-08-19-owner-problem-fixes-overview.md`.

## Problem

An invite whose recipient address already belongs to a registered identity was only rejected **at redemption time**. The invitee filled in the whole registration, signed and forwarded the contract, and only then hit the dead-end "invitation already used" page. All of that work wasted.

## Root cause

`createInvite` performed no availability check on the recipient address. The collision was only discovered later, at redemption, by which point the invitee had already done the work — and the admin who sent the invite never learned it was doomed.

## Fix

`createInvite` now verifies the recipient address against the existing `IdentityEmailOwnerLookup` port **before anything is reserved or persisted**, and answers **409 with `X-Reason: EMAIL_NOT_AVAILABLE`**. That reason makes the failure distinguishable from the bare 400 of a malformed address and from the reason-less 409 of a taken tenant ID.

Two properties that matter for review:

- **The guard sits before the tenant-ID reservation.** One guard therefore covers send-directly and add-to-list, tenant and counsellor invites, and CSV import — they all funnel through this one call. It is role-agnostic by construction, not by enumeration.
- **No new endpoint is introduced.** The check rides on the existing admin-only create call, so it never becomes an unauthenticated **user-enumeration oracle**, and there is no TOCTOU gap between a preflight answer and the real create.

The address is normalized (trim + lower-case) exactly the way `CounsellorInviteProvisioningService` normalizes it when it later creates the identity, so the guard tests the value that would actually collide.

## Tests

4 new `@Test` cases in `AccountInviteServiceTest` (+101 lines): address already owned → 409 + reason and **nothing reserved**; address free → unchanged happy path; normalization (whitespace / mixed case) collides correctly; the reservation is genuinely not touched on refusal.

## Reviewer test plan

- [ ] Create an invite for an address that already belongs to a registered identity — **409** with `X-Reason: EMAIL_NOT_AVAILABLE`.
- [ ] Confirm **no tenant-ID reservation** was consumed by the refused call.
- [ ] Same address with leading/trailing whitespace and different casing — still refused (normalization matches `CounsellorInviteProvisioningService`).
- [ ] Counsellor invite and tenant-admin invite — both refused by the same guard.
- [ ] „Direkt Versenden" (with `templateId`) and „Empfänger nur anlegen" (without) — both refused.
- [ ] CSV import with one colliding row — that row carries the reason, the others still import.
- [ ] Tenant-ID collision — still the reason-**less** 409, distinguishable from this one.
- [ ] Malformed address — still 400, not 409.
- [ ] Confirm **no new route** appears in the OpenAPI surface (no enumeration oracle).
- [ ] **Mobile (390 px)**: with the Admin branch deployed, the refusal renders inline on the e-mail field on a phone viewport.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/keycloak/KeycloakService.java
@@ -226,10 +226,20 @@ public class KeycloakService
     updateEmail(userId, userHelper.getDummyEmail(userId));
   }
 
+  /**
+   * Exact-owner lookup on top of Keycloak's fuzzy user search, which also matches on username,
+   * first and last name — hence the re-filter on the e-mail field itself.
+   *
+   * <p>The comparison ignores case: callers normalize the probe to lower case, but a stored record
+   * need not be lower-cased (imported or externally federated users routinely are not). A
+   * case-sensitive comparison would discard exactly the hit that Keycloak's own case-insensitive
+   * search just returned and report the address as free — the same duplicate-address defect the
+   * callers use this method to prevent.
+   */
   @Override
   public Optional<IdentityEmailOwner> findByEmail(String email) {
     return keycloakClient.getUsersResource().search(email, 0, Integer.MAX_VALUE).stream()
-        .filter(userRepresentation -> email.equals(userRepresentation.getEmail()))
+        .filter(userRepresentation -> email.equalsIgnoreCase(userRepresentation.getEmail()))
         .findFirst()
         .map(userRepresentation -> new IdentityEmailOwner(userRepresentation.getUsername()));
   }
--- a/src/main/java/de/caritas/cob/userservice/api/port/out/AccountInviteRepository.java
+++ b/src/main/java/de/caritas/cob/userservice/api/port/out/AccountInviteRepository.java
@@ -61,6 +61,49 @@ public interface AccountInviteRepository extends JpaRepository<AccountInvite, Lo
   boolean existsByTenantIdAndTargetRoleAndStatusIn(
       Long tenantId, AccountInviteTargetRole targetRole, Collection<AccountInviteStatus> statuses);
 
+  /**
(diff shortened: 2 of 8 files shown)


kiodreambau:
### Why a test-only commit appeared here (a1b0d84f)

This branch was red on the `required integration tests` gate, and it is not a merge artefact — it reproduces on its own on top of clean `pre-dev`.

Adding the duplicate-e-mail guard made `IdentityEmailOwnerLookup` the 11th constructor dependency of `AccountInviteService`. Two pre-existing `@DataJpaTest` slices build that bean with `@Import(AccountInviteService.class)` and declare exactly one `@MockitoBean` per collaborator, so a slice with no matching mock can no longer create its context:

```
UnsatisfiedDependencyException: Error creating bean with name
'...accountinvite.AccountInviteService': Unsatisfied dependency expressed through
constructor parameter 10: No qualifying bean of type
'...port.out.IdentityEmailOwnerLookup' available
```

That failed `AccountInviteAcceptRaceIT` (2 tests) and `AccountInviteReservationOrchestrationIT` (3 tests) — 5 errors out of 895, only the first being a real context-load failure and the rest the `ApplicationContext failure threshold (1) exceeded` cascade.

The fix is additive only: one import and one `@MockitoBean` per file, 4 insertions, 0 deletions, no existing mock touched. The guard itself is unchanged and the new dependency stays mandatory in production code. The port is consulted only on the creation path and Mockito answers `Optional.empty()` by default, so neither slice changes what it asserts.

The same commit is also carried on the throwaway integration branch `speed/owner-fixes-2026-08-19` (as `37d30977`); the two are byte-identical — cherry-picked with `-x`, and both file blobs hash the same.

shazia-k (approved):

PR #1074, opened 2026-08-24
Join late-joining counsellors into existing enquiry rooms (and revoke on removal)

kiodreambau:
Closes OpenResilienceInitiative/ORISO-UserService#1060
Refs OpenResilienceInitiative/ORISO-Frontend#811

## Why this exists twice

PR OpenResilienceInitiative/ORISO-UserService#1063 already implemented this issue — but its base was `speed/reflux-progress-exp-16-aug`, not `pre-dev`. That branch is stuck behind a red openapi consumer check, so the fix never reached `pre-dev` and the bug is still live there:

```
git ls-tree -r --name-only origin/pre-dev | grep -i latejoiner   → empty
```

This PR brings the work onto `pre-dev` and closes four gaps #1063 left open (integration test, rollback leak, `INITIAL` enquiries, bulk-import fan-out).

## Problem

US#905 joins the agency's counsellors into an enquiry's Matrix room **at room creation**. That only helps counsellors who were already in the agency at that moment. A counsellor added afterwards never becomes a member of the rooms that already exist — Matrix `/sync` returns nothing for a non-member, and no later event repairs it. The agency's open initial requests simply stay invisible to them: no error, no empty state, nothing to act on.

## Change

`AgencyLateJoinerMembershipService` closes both halves:

- **Join (the reported bug):** completing an agency assignment backfills the counsellor into the Matrix rooms of the agency's existing open enquiries.
- **Revoke (the reverse path the issue asks for):** marking a consultant-agency relation deleted takes that membership back, so a counsellor does not keep syncing enquiries of an agency they left. The hook sits in `markAsDeleted`, which covers both the single-agency endpoint and the bulk path behind `ConsultantAdminFacade#setConsultantAgencies`.

Per-counsellor membership logic is shared with the at-creation fan-out by promoting `AgencySilentMembershipService#joinSilently` to the public `joinConsultantIntoRoom`, so the backfill cannot drift from #905's semantics.

**Scope:** open enquiries — status `NEW` or `INITIAL`, no counsellor assigned, not directly addressed. Those are the cases the whole department is entitled to pick up. `INITIAL` is more than extra coverage: a counsellor backfilled before the first message is a member when the asker sends, so the Megolm session reaches their device and they read plaintext instead of depending on the client-side history-key transfer. On the removal side the scope is a constraint rather than a preference — `AssignEnquiryFacade` removes the agency service account from the room when an enquiry is accepted, so its token can no longer act on an accepted case.

## The rollback leak #1063 left open

`ConsultantAdminFacade#setConsultantAgencies` is `@Transactional` and applies removals before creations, one agency at a time. #1063 called the fan-out inline. If a later agency is rejected, the first agency's Matrix join has **already happened** and the database rolls back around it — the counsellor ends up a member of the enquiry rooms of an agency they were never given. The mirror case on the removal side locks them out of enquiries the database still says are theirs.

Fixed by publishing `ConsultantJoinedAgencyEvent` / `ConsultantLeftAgencyEvent` and acting on them in `AgencyMembershipSyncListener` with `@TransactionalEventListener(AFTER_COMMIT, fallbackExecution = true)`. `fallbackExecution` is load-bearing, not decorative: the POST endpoint and the CSV import are not transactional at all, and without it the listener would silently never run on those paths.

This is verified, not asserted. Temporarily downgrading the listener to a plain synchronous `@EventListener` (exactly #1063's semantics) makes the guard test fail:

```
[ERROR] ConsultantAgencyLateJoinerMembershipIT
        .replacingAgencies_joinsNoRoom_When_aLaterAssignmentInTheSameTransactionFails
No interactions wanted here
```

Restoring `AFTER_COMMIT` makes it pass.

## Failure semantics

Both directions are best effort. The relation in the database is the source of truth, so a Synapse failure degrades to "membership repaired later" rather than failing the admin's request. Nothing escapes the listener — `CreateConsultantSaga#assignAgenciesOrRollback` deletes the freshly created consultant on any `RuntimeException`, so an unhandled Matrix error there would cost an admin the consultant they just created.

## Bulk import

`ConsultantImportService` creates relations row by row and agency by agency, and the fan-out costs Synapse calls per open enquiry room. A large import multiplies that out. `matrix.membership.lateJoiner.enabled` (`MATRIX_LATE_JOINER_MEMBERSHIP_ENABLED`, default `true`) lets operations switch the fan-out off for an import window; membership is repaired on the next assignment. No Helm change is needed for the default.

The service also queries and filters sessions **before** touching the agency service or Matrix, and returns early when the agency has no open enquiry room. That ordering is what keeps the six existing consultant-agency ITs from making real outbound calls.

## Tests

Written test-first; every test was watched failing before the code that makes it pass existed on this branch.

- `ConsultantAgencyLateJoinerMembershipIT` (new, 4 tests) — drives the real admin entry points end to end. Before the fix, on clean `pre-dev`:
  ```
  Wanted but not invoked:
  matrixSessionRoomGateway.joinRoom("!existing-enquiry:oriso.org", "late-joiner-token");
  Actually, there were zero interactions with this mock.
  ```
  Covers the POST path (no outer transaction → exercises `fallbackExecution`), the PUT path the admin panel actually calls (real `AFTER_COMMIT`), agency isolation, and the rollback guard.
- `AgencyLateJoinerMembershipServiceTest` (new, 11) — join and revoke fan-out, directly-addressed exclusion, rooms-less enquiries, missing agency service account, per-room best effort, `INITIAL` coverage, and the "never touches Matrix when there is nothing to do" guard.
- `AgencyMembershipSyncListenerTest` (new, 7) — resolution by id, never-throws in both directions, consultant deleted between commit and fan-out, kill switch.
- `ConsultantAgencyRelationCreatorServiceTest` (+3) — publishes on both the atomic and the legacy two-step path, ordered after the finalizer, and stays silent when the relation is rejected.
- `ConsultantAgencyAdminServiceTest` (+3) — publishes after the relation is persisted as deleted, stays silent when the deletion is refused, one event per agency in the bulk path.

Verification run locally (JDK 21):

```
./mvnw -B test                                    → 4020 tests, 0 failures
./mvnw -B -Dskip.unit-tests=true -Dtest=…IT …     → 40 tests, 0 failures
python3 scripts/ci/check-no-test-quarantine.py    → OK
python3 -m unittest discover -s tests/ci          → 97 tests, OK
```

`AgencyPreAssignmentRoomService` and its 16 tests are deliberately **not** touched — that is the evidence for the issue's second acceptance criterion ("existing at-creation behavior from #905 unchanged").

## Known limits (deliberate, recorded on the issue)

- **Megolm.** For messages sent *before* the backfill the backend can deliver membership and ciphertext, not plaintext. Reading those depends on the client-side history-key transfer from ORISO-Frontend#811. Enquiries backfilled while still `INITIAL` do not have this problem at all.
- **Revoke scope.** Only open enquiries. A counsellor who leaves an agency keeps membership in rooms of enquiries that were already accepted, because the agency service account has left those rooms and its token cannot kick anyone out of them.

## How should I test this?

- [ ] Create an enquiry for agency A as an advice seeker → confirm a counsellor already in agency A sees it → **expected:** unchanged behaviour from #905, enquiry visible and readable.
- [ ] Add a **new** counsellor to agency A in the admin panel → log in as that counsellor → open the initial-request list → **expected:** the enquiry created *before* they joined is listed and opens.
- [ ] Same enquiry, check attachments → **expected:** attachments load. If the message text stays unreadable, that is the Megolm limit above, not this hook — check the Synapse membership dump to confirm the join itself succeeded.
- [ ] Remove that counsellor from agency A again → **expected:** the enquiry disappears from their list.
- [ ] Service log during the assignment → **expected:** one `Backfilled consultant … into N of M open enquiry rooms of agency …` line.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/admin/service/agency/ConsultantAgencyAdminService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/admin/service/agency/ConsultantAgencyAdminService.java
@@ -22,18 +22,22 @@ import de.caritas.cob.userservice.api.port.out.ConsultantAgencyRepository;
 import de.caritas.cob.userservice.api.port.out.ConsultantRepository;
 import de.caritas.cob.userservice.api.port.out.SessionRepository;
 import de.caritas.cob.userservice.api.service.agency.AgencyService;
+import de.caritas.cob.userservice.api.service.session.ConsultantLeftAgencyEvent;
 import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 import lombok.NonNull;
 import lombok.RequiredArgsConstructor;
 import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.beanutils.BeanUtils;
+import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.stereotype.Service;
 
 /** Service class to handle administrative operations on consultant-agencies. */
 @Service
 @RequiredArgsConstructor
+@Slf4j
 public class ConsultantAgencyAdminService {
 
   private final @NonNull ConsultantAgencyRepository consultantAgencyRepository;
@@ -43,6 +47,7 @@ public class ConsultantAgencyAdminService {
   private final @NonNull AgencyService agencyService;
   private final @NonNull AgencyAdminService agencyAdminService;
   private final @NonNull ConsultantAgencyDeletionValidationService agencyDeletionValidationService;
+  private final @NonNull ApplicationEventPublisher eventPublisher;
 
   /**
    * Returns all Agencies for the given consultantId.
@@ -258,6 +263,33 @@ public class ConsultantAgencyAdminService {
     this.agencyDeletionValidationService.validateAndMarkForDeletion(consultantAgency);
     consultantAgency.setDeleteDate(nowInUtc());
     this.consultantAgencyRepository.save(consultantAgency);
+    announceDetachedAgency(consultantAgency);
+  }
+
+  /**
+   * US#1060, symmetric half: a counsellor is joined into the agency's open enquiry rooms when they
+   * are assigned, so detaching them again has to take that membership back. Otherwise they keep
+   * receiving the agency's open initial requests through Matrix {@code /sync} long after the
+   * relation was deleted.
+   *
+   * <p>Announced rather than performed here, and only once the deletion is persisted: the bulk path
+   * behind {@code ConsultantAdminFacade#setConsultantAgencies} applies removals inside a
+   * transaction that a later rejected assignment can still roll back. Revoking Matrix membership
+   * inline would lock the counsellor out of enquiries the database still says are theirs. See
+   * {@code AgencyMembershipSyncListener}.
+   */
+  private void announceDetachedAgency(ConsultantAgency consultantAgency) {
+    var consultant = consultantAgency.getConsultant();
+    if (consultant == null) {
+      log.warn(
+          "Consultant agency relation {} has no consultant; enquiry room membership of agency {}"
+              + " not revoked",
+          consultantAgency.getId(),
+          consultantAgency.getAgencyId());
+      return;
+    }
+    eventPublisher.publishEvent(
+        new ConsultantLeftAgencyEvent(consultant.getId(), consultantAgency.getAgencyId()));
   }
 
   /**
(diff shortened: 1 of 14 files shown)


kiodreambau:
### CI note for the reviewer: one flaky job, not a regression

The first run of `mariadb-contract` failed on `TutorialProgressServiceMariaDbReplicaIT.concurrentNewScopesForOneUserNeverExceedTheRowCap` (`expected: 1 but was: 2`). A rerun with **no code change** went green (4m26s), and all required checks now pass.

It is a flake in that test, unrelated to this PR:

- The **same commit** produced a `pass` (4m0s) and a `fail` (2m14s) across the two workflow runs.
- The test coordinates two threads over a hand-built `TutorialProgressStore` against a real MariaDB container; this PR touches consultant-agency assignment and Matrix membership. No shared bean, no shared state.
- The assertion sits behind `secondCountReached.await(500, TimeUnit.MILLISECONDS)` — on a loaded runner the second thread reaches `count()` before the first is paused, so the counter reads 2 instead of 1. The 500 ms window is the race.

Flagging it rather than silently rerunning: the test is timing-dependent and will keep failing intermittently on `pre-dev` until that latch window is replaced with a deterministic handshake. Worth a look independently of this PR.

shazia-k (approved):

PR #1102, opened 2026-09-02
Fix live chat session data lazy loading

Hassan9215:
## Summary
- Add read-only transactions around cross-tenant live-chat session open fallbacks so DTO mapping can read lazy sessionData safely.
- Add regression coverage that keeps both fallback methods transaction-scoped.

## Context
Dev logs showed LazyInitializationException while opening anonymous LIVE_CHAT sessions through UserSessionControllerDelegate -> SessionService.getVisibleAnonymousLiveChatEnquiriesByIds -> SessionMapper.buildSessionDataMapFromSession. The same vulnerable code path exists on pre-dev even though pre-dev data does not currently trigger it.

## Test
- ./mvnw -Dtest=SessionServiceTest test

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/service/session/SessionService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/session/SessionService.java
@@ -657,6 +657,7 @@ public class SessionService {
    * fallback the open path uses so a visible request can always be opened; it never widens access
    * to registered sessions.
    */
+  @Transactional(readOnly = true)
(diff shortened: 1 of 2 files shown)


Hassan9215:
Reviewer note: I attempted to request review from @chatzi, but GitHub rejected it because that account is not a collaborator on OpenResilienceInitiative/ORISO-UserService. Please add the correct collaborator/team reviewer if Chatzi uses a different GitHub login.

shazia-k (approved):

PR #1110, opened 2026-09-04
fix(absence): let the public consultant endpoint carry the out-of-office message

shazia-k:
> **This PR now carries two unrelated fixes.** It began as the out-of-office fix below; the tenant-context leak was added at the author's request to unblock `required integration tests`, which was failing on every branch. The two are in separate commits and can be reviewed independently.

## What

`GET /users/consultants/{consultantId}` (unauthenticated) and its public-slug twin never returned the counsellor's absence state. `ConsultantResponseDTO` had no `absent`/`absenceMessage` properties at all, and `ConsultantDtoMapper.consultantResponseDtoOf` never set them.

The frontend has rendered an absence notice from exactly those two fields for a while — `PreselectedConsultant.tsx:30` and `preselectionDrawer.tsx:29` both branch on `consultant?.absent` and print `consultant.absenceMessage`. It typechecks because `ConsultantDataInterface` inherits both as optional from `UserDataInterface`, so `absent` was silently `undefined` forever and the branch never taken.

Net effect: an advice seeker following a counsellor's link was never told the counsellor was away, no matter what that counsellor had saved under "Meine Abwesenheit".

## Changes

- `api/userservice.yaml` — add `absent` and `absenceMessage` to `ConsultantResponseDTO`.
- `ConsultantDtoMapper.consultantResponseDtoOf` — populate both. `absenceMessage` is returned **only while `absent` is true**: this endpoint is unauthenticated, and a stale message from a counsellor who has already returned has no business being publicly readable.
- Two tests in `ConsultantDtoMapperTest` covering the absent and not-absent cases.

No frontend change is needed — the UI is already there and reads `absent` / `absenceMessage`.

## Testing

`ConsultantDtoMapperTest` — 24/24 pass.

## Scope note

This does **not** touch the in-chat composer banner, the other surface where absence is shown. I checked that path and could not fault it on `dev`:

- `SessionMapper.toGroupSessionResponse` maps absence into the room-session response on `dev`.
- Serializing the compiled `GroupSessionConsultantDTO` / `SessionConsultantForUserDTO` from `target/UserService.jar` with the bundled Jackson 3.1.4 yields `{"absenceMessage":"...","displayName":"...","absent":true}`.
- The deployed dev bundle reads `ASKER_DEFAULT && consultant?.absent` and renders `consultant.absent.message` + `absenceMessage`, matching source.

If a blank composer is still seen in a *running* chat, that needs a live capture of `/service/users/sessions/room` for the affected asker — it is not reproducible from the code.

## Follow-up (not in this PR)

`api/userservice.yaml` documents the same flag as `isAbsent` on `SessionConsultantForUserDTO` and `GroupSessionConsultantDTO`, but the server sends `absent` (Lombok `isAbsent()` + Jackson bean naming). `ORISO-Frontend/src/generated/userservice.d.ts` is therefore wrong for that field; runtime code is unaffected only because it reads the hand-written `SessionConsultantInterface`.

Refs OpenResilienceInitiative/ORISO-Frontend#1210

## Summary by CodeRabbit

* **New Features**
  * Consultant responses now indicate whether a consultant is currently absent.
  * Absence messages are included only when the consultant is absent.

* **Bug Fixes**
  * Availability information is represented consistently in consultant responses.
  * Request-specific tenant information is cleared after each request, preventing it from affecting subsequent requests.

---

# Second fix: registration leaks its tenant onto the thread

Unrelated to the absence work, added here to unblock CI.

## The bug

`/users/askers/new` is whitelisted in `HttpTenantFilter`, so a registration is never wrapped by the filter that tears the tenant context down. `CreateUserFacade.createUserAccountWithInitializedConsultingType` calls `initializeTenantContextForRegistration`, which sets `TenantContext` from the agency and never clears it. `TenantContext` is a ThreadLocal on a pooled thread, so the tenant outlives the request and scopes whatever lands on that thread next.

**This is a production bug, not only a test-ordering one.** In a real container those threads are shared across users, so a registration can leave a tenant id that silently scopes an unrelated later request — cross-tenant scoping on a multi-tenant platform.

## Why CI looked flaky

`UserControllerE2EIT` registers seven times and leaks a tenant. Whether anything breaks depends purely on run order:

| | `UserControllerE2EIT` | failing classes | Result |
|---|---|---|---|
| Linux (CI) | 70 | 71, 73 | fails — leak lands first |
| macOS (local) | 46 | 31 | passes — leak lands after |

Downstream, `AccountManager.resolveEffectiveTenantId` falls back to the ThreadLocal and `ConsultantRepository.findAllByInfix` filters on `(?2 IS NULL OR ?2 = 0 OR c.tenantId = ?2)`, so a leaked id matches nothing and the search returns `total: 0`.

Instrumenting a replay of CI's order caught it directly: `effectiveTenantId=1326634973105178603` — the EasyRandom value from the mocked agency, still on the thread.

## Changes

`9bd6b32` adds `TenantContextCleanupFilter`: an unconditional `OncePerRequestFilter` at highest precedence that clears the ThreadLocal in a `finally`.

Cleaning up at the request boundary rather than inside the handler is the load-bearing part. An earlier attempt cleared inside `CreateUserFacade` and **broke registration**: `UserRegistrationControllerDelegate#registerUser` calls `markAsDirectConsultant` after the facade returns, and with multitenancy on a tenant-filtered query under a cleared context matches nothing rather than failing — measured at 100 rows under the technical context versus 0 under a cleared one — so the session would not be found and a registration with a preselected counsellor would answer 500. No suite caught that: the `testing` profile disables multitenancy, so `TenantAspect` is never registered. That approach was dropped.

The filter is unconditional on purpose, unlike `HttpTenantFilter` which only exists when `multitenancy.enabled`. The ThreadLocal can be set on any profile, so it has to be cleaned on any profile.

## Verification

Replaying CI's class order locally reproduces all 6 `searchConsultants` failures plus `TutorialStatisticsControllerE2EIT` before the fix, and none after. Full integration suite **908/908**. Unit suite 4243/4245 — the 2 failures are `MatrixOnlyRuntimeConfigurationContractTest` asserting on a gitignored `application-local.properties` that exists only on my machine and never reaches CI.

### DIFF
--- a/api/userservice.yaml
+++ b/api/userservice.yaml
@@ -2901,7 +2901,7 @@ components:
         absenceMessage:
           type: string
           example: I am absent until...
-          description: The out-of-office message; only present when absent is true, otherwise omitted
+          description: The out-of-office message; carries a value only while absent is true, otherwise null
         agencies:
           type: array
           items:
--- a/api/userservice.yaml
+++ b/api/userservice.yaml
@@ -2900,8 +2900,9 @@ components:
           description: Flag that indicates if the consultant is currently absent
         absenceMessage:
           type: string
+          nullable: true
           example: I am absent until...
-          description: The out-of-office message the consultant published for their absence
+          description: The out-of-office message; only returned when absent is true, otherwise null/omitted
         agencies:
           type: array
           items:
--- a/api/userservice.yaml
+++ b/api/userservice.yaml
@@ -2900,6 +2900,7 @@ components:
           description: Flag that indicates if the consultant is currently absent
         absenceMessage:
           type: string
+          nullable: true
           example: I am absent until...
           description: The out-of-office message; carries a value only while absent is true, otherwise null
         agencies:
--- a/api/userservice.yaml
+++ b/api/userservice.yaml
@@ -2900,9 +2900,8 @@ components:
           description: Flag that indicates if the consultant is currently absent
         absenceMessage:
           type: string
-          nullable: true
           example: I am absent until...
-          description: The out-of-office message; only returned when absent is true, otherwise null/omitted
+          description: The out-of-office message; only present when absent is true, otherwise omitted
         agencies:
           type: array
           items:
--- a/api/userservice.yaml
+++ b/api/userservice.yaml
@@ -2894,6 +2894,14 @@ components:
           type: boolean
           example: true
           description: Flag that indicates if the consultant can be added as a supervisor
+        absent:
+          type: boolean
+          example: true
+          description: Flag that indicates if the consultant is currently absent
+        absenceMessage:
+          type: string
+          example: I am absent until...
+          description: The out-of-office message the consultant published for their absence
         agencies:
           type: array
           items:
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/interceptor/TenantContextCleanupFilter.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/interceptor/TenantContextCleanupFilter.java
@@ -43,6 +43,7 @@ public class TenantContextCleanupFilter extends OncePerRequestFilter {
   protected void doFilterInternal(
       HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
       throws ServletException, IOException {
+    TenantContext.clear();
     try {
       filterChain.doFilter(request, response);
     } finally {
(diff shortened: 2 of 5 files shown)


Storypapst:
@shazia-k please make sure all issues are correctly linked, in the right version and progress lane.

Shirloin:
Branch validation's IT check is red, but it looks flaky, not cause by this PR

Shirloin (dismissed):
CI now green

Shirloin (dismissed):

Shirloin (approved):

PR #1138, opened 2026-09-10
fix(security): permit anonymous reads on the service-prefixed consultant-languages route

### LINKED ISSUE #1119: Public registration receives 401 from service-prefixed consultant languages endpoint
## Why

Anonymous registration on Dev calls `GET /service/users/consultants/languages?agencyId=<id>`, but UserService returns `401 Unauthorized`. The public security allow-list currently covers only `/users/consultants/languages`, while the deployed frontend uses the service-prefixed route. As a result, registration cards can show an empty language list even though consultant languages exist.

## What to build

- Permit anonymous reads for the service-prefixed consultant-languages endpoint used by registration.
- Preserve the existing unprefixed public route and keep unrelated UserService endpoints protected.
- Add a regression test that performs the request without an authenticated user.

## Acceptance criteria

- [ ] Anonymous `GET /service/users/consultants/languages?agencyId=<valid-id>` returns `200` and the expected language payload.
- [ ] The existing public `/users/consultants/languages` behavior remains unchanged.
- [ ] Registration on Dev renders the returned consultant languages instead of failing on `401`.
- [ ] No broader anonymous access is introduced for UserService routes.

## Reviewer test plan

- [ ] Run the focused anonymous endpoint integration test.
- [ ] Run the UserService quality gate for the changed module.
- [ ] In a fresh browser session on Dev, open registration for an agency with consultant languages and verify the network request returns `200`.
- [ ] Confirm another protected UserService endpoint still rejects an anonymous request.

shazia-k:
Closes #1119

## Problem

Anonymous registration calls `GET /service/users/consultants/languages?agencyId=<id>` and the API gateway forwards `/service` unchanged. `SecurityConfig` permitted only `/users/consultants/languages`, so the request fell through to the authenticated matcher and returned **401**. Registration cards then rendered an empty language list even for agencies that do have consultant languages.

There are two gaps, not one. Permitting the matcher alone would have turned the 401 into a **404**, because the generated mapping on `UserController.getLanguages` only carries the unprefixed path.

## Change

- `SecurityConfig` — `/service/users/consultants/languages` added to the public allow-list, next to the existing unprefixed entry.
- `UserController.getLanguages` — dual `@GetMapping` over both paths, with `@RequestParam` re-declared because a method-level mapping does not inherit the interface's parameter annotations. Same shape as `registerUser` and `getSessionForId`, which already handle the gateway prefix this way.

No OpenAPI change: the spec never lists `/service` variants, so the generated-diff check is unaffected.

## Tests

Three cases in `UserControllerAuthorizationIT`:

- anonymous `GET /service/users/consultants/languages?agencyId=1` → 200, payload asserted, delegation verified;
- anonymous `GET /users/consultants/languages?agencyId=1` → 200, existing behaviour unchanged;
- anonymous `GET /service/users/sessions/room/1` → 401, so the `/service` prefix does not become blanket-public.

## Notes for review

- `StatelessCsrfFilter` exempts GET, so the anonymous reads need no CSRF token.
- `multitenancy.enabled` is `false` in every environment, so `HttpTenantFilter` is not registered. Were it enabled, its whitelist matches on `requestUri.contains(...)`, which already covers `/service` prefixes.
- The three existing `getLanguages` E2E cases are untouched: they use the unprefixed path with `@WithMockUser`, and the missing-`agencyId` case still returns 400 through `required = true`.

## Reviewer test plan

- [ ] `mvn -Dtest=UserControllerAuthorizationIT test`
- [ ] UserService quality gate for the changed module
- [ ] On Dev, in a fresh browser session, open registration for an agency with consultant languages and confirm the request returns 200 with the languages rendered
- [ ] Confirm another protected UserService endpoint still rejects an anonymous request

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving consultant languages through both standard and service-prefixed API paths.
  * Consultant language requests now accept a required agency identifier and return JSON results.
  * These language endpoints are accessible without authentication.

* **Bug Fixes**
  * Preserved authentication requirements for user session room endpoints.

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserController.java
+++ b/src/main/java/de/caritas/cob/userservice/api/adapters/web/controller/UserController.java
@@ -331,8 +331,22 @@ public class UserController implements UsersApi {
     return userAccountControllerDelegate.updateConsultantData(updateConsultantDTO);
   }
 
+  /**
+   * Returns the languages of all consultants working for the given agency.
+   *
+   * <p>Public by design: anonymous registration reads this list before an account exists. Mapped to
+   * both the direct path and the /service/ prefix because the API gateway forwards /service
+   * unchanged and the deployed frontend calls the prefixed route.
+   *
+   * @param agencyId (required) the agency to read the consultant languages of
+   * @return {@link ResponseEntity} containing {@link LanguageResponseDTO}
+   */
+  @GetMapping(
+      value = {"/users/consultants/languages", "/service/users/consultants/languages"},
+      produces = MediaType.APPLICATION_JSON_VALUE)
   @Override
-  public ResponseEntity<LanguageResponseDTO> getLanguages(Long agencyId) {
+  public ResponseEntity<LanguageResponseDTO> getLanguages(
+      @RequestParam(value = "agencyId", required = true) Long agencyId) {
     return userConsultantControllerDelegate.getLanguages(agencyId);
   }
 
--- a/src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
+++ b/src/main/java/de/caritas/cob/userservice/api/config/auth/SecurityConfig.java
@@ -136,6 +136,7 @@ public class SecurityConfig {
                     "/service/conversations/anonymous/availability",
                     "/users/consultants/{consultantId:" + PUBLIC_CONSULTANT_ID_PATTERN + "}",
                     "/users/consultants/languages",
+                    "/service/users/consultants/languages",
                     "/error-reports",
                     "/service/error-reports",
                     "/users/magic-link/request",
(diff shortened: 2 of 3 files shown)


Shirloin (approved):
Approved. Anonymous `GET /service/users/consultants/languages?agencyId=2` returns 401 on Dev today while the unprefixed path returns 200 — so registration cards show an empty language list, and someone looking for counselling in their own language can't tell whether the agency offers it or the request just failed.

The change opens that one path, not the `/service` prefix as a whole, and the suite pins it: a protected route on the same prefix still rejects anonymous callers. The data is already public through the unprefixed route, so nothing new is exposed.

Not yet confirmed: the 200 after the fix, since this branch isn't on Dev. Also, `closingIssuesReferences` is empty despite the body saying "Closes #1119" — #1119 may need closing by hand.

PR #1139, opened 2026-09-10
fix(invites): release the tenant-ID reservation when an invite dies

### LINKED ISSUE #1052: Revoking a tenant-admin invite leaks its tenant-ID reservation
## Problem

Revoking a tenant-admin invite does not release the tenant ID it reserved. The reservation stays
`RESERVED` in TenantService forever, so every revoked — and presumably every expired — invite
permanently burns one tenant ID from a small, human-meaningful ID space.

## Actual behavior

Observed on pre-dev while building the AVV E2E spec
(OpenResilienceInitiative/ORISO-E2E#67). Three invites were created in AUTO mode and then revoked
through `POST /service/useradmin/account-invites/{id}/revoke`, which returned `200`:

```
invites after revoke:  15 (tenantId 19, REVOKED)
                       14 (tenantId 18, REVOKED)
                       13 (tenantId 17, REVOKED)

GET /service/tenantadmin/tenant-ids/17/availability -> {"id":17,"status":"RESERVED"}
GET /service/tenantadmin/tenant-ids/18/availability -> {"id":18,"status":"RESERVED"}
GET /service/tenantadmin/tenant-ids/19/availability -> {"id":19,"status":"RESERVED"}
```

The IDs only became free after an explicit
`DELETE /service/tenantadmin/tenant-ids/reservations/{id}` (each `204`):

```
GET .../17/availability -> {"id":17,"status":"FREE"}
GET .../18/availability -> {"id":18,"status":"FREE"}
GET .../19/availability -> {"id":19,"status":"FREE"}
```

## Expected behavior

Revoking (or expiring) a tenant-admin invite releases its tenant-ID reservation, so the ID returns
to `FREE` and can be allocated again.

## Why this matters

`AccountInviteService` already has exactly this compensation logic for the *creation* failure path:

```java
} catch (RuntimeException exception) {
  // Compensation: a failed creation must not leave orphaned reservations behind.
  if (reservedAgencyId != null) { agencyIdAllocationClient.release(reservedAgencyId); }
  if (tenantReservation != null) { tenantIdAllocationClient.release(tenantReservation.tenantId()); }
  throw exception;
}
```

The same reasoning applies to revoke and to expiry — the reservation outliving the invite that
justified it is the orphan this comment is about. Agency IDs are reserved the same way
(`agencyIdAllocationClient`) and are worth checking for the identical gap.

Tenant IDs are short, admin-visible and chosen deliberately (the Admin panel exposes an "Auto /
next free" picker), so silently consuming them on every abandoned invite degrades the ID space
over time and makes the "next free ID" suggestion drift away from reality.

## Evidence

Run id `e2e-20260817-0430`. The E2E spec added in OpenResilienceInitiative/ORISO-E2E#67 works
around this by releasing the reservation explicitly in its cleanup, with a comment pointing here.

shazia-k:
Closes #1052

## Problem

Revoking a tenant-admin invite did not release the tenant ID it reserved. The reservation stayed `RESERVED` in TenantService forever, so every revoked — and every expired — invite permanently burned one ID from a small, human-meaningful ID space. Because the Admin panel offers a "next free ID" picker, that suggestion drifts away from reality as the space degrades.

`createInvite` already carries exactly this compensation on its failure path. Revoke and expiry are the same orphan.

## Change

`AccountInviteService` gains `releaseTenantIdReservationAfterCommit`, called from `revokeInvite` and from the tenant-admin expiry transition in `TenantAdminOnboardingService`.

**After commit, not inline.** The reservation may only be freed once the terminal status is durable — releasing inside the transaction would hand the ID back and then, on a rollback, leave a live invite pointing at an ID somebody else can already take. This uses the same `TransactionSynchronization` idiom as `CaseHandoverService.attachStandingSupervisorAfterCommit`. Outside a transaction it runs inline, which carries the same guarantee because `release` never throws.

**Nothing escapes `afterCommit`.** The invite is durably dead by then, and an escaping exception would surface as a 500 on a revoke that has in fact succeeded. The client swallows its own transport errors; the guard covers the rest (header supply, client construction).

## Three cases that must not release

| Case | Why |
|---|---|
| No reservation held (`tenantIdReservationToken == null`) | Nothing to give back |
| `SUPERSEDED` invite | `resendInvite` hands the very same reservation to the replacement; releasing would pull the ID out from under a live invite |
| Second revoke of an already-`REVOKED` invite | The ID may have been re-reserved by someone else in the meantime |

## Deliberately out of scope

**Agency IDs.** The issue asks whether the identical gap exists there. It does not have a safe fix yet: `AccountInvite.agencyId` carries either a reserved ID or a pre-existing agency's ID, and the row does not record which, so releasing it here could free an ID belonging to a live agency. Closing that half needs the allocation mode persisted on the invite first. Pinned by a test so the boundary is explicit rather than accidental.

**The expiry transition in `AccountInviteService.acceptInvite`.** Not wired, because it cannot fire: `AccountInviteLinkException extends RuntimeException` and the method is `@Transactional` without `noRollbackFor`, so the `save(EXPIRED)` immediately before the throw is rolled back with it. `TenantAdminOnboardingService.expireIfPastExpiry` returns the link-death answer instead of throwing precisely so its write commits, which is why the release is wired there and only there. Worth a separate look — the `EXPIRED` write in `acceptInvite` appears to be lost today.

## Tests

`AccountInviteServiceTest` — releases when a reservation is held; does not release without a token, for a superseded invite, or on a second revoke; never touches the agency ledger.

`AccountInviteReservationOrchestrationIT` — `revokedInvite_Should_LeaveNoReservationBehind` mirrors the PreDev evidence in the issue: the in-memory ledger is empty again and the row is `REVOKED`. A sibling pins the agency-ledger boundary.

`OnboardingInviteExpiryCommitIT` — asserts the release on both tenant-admin expiry paths. This IT previously did not mock `TenantIdAllocationClient` while seeding an invite with `tenantId=79` and a reservation token, so the new call would have reached the real client; it is now a `@MockitoBean`.

## Reviewer test plan

- [ ] `mvn -Dtest='AccountInviteServiceTest,AccountInviteReservationOrchestrationIT,OnboardingInviteExpiryCommitIT' test`
- [ ] UserService quality gate for the changed module
- [ ] On PreDev, repeat the issue's steps: create an AUTO-mode tenant-admin invite, revoke it, then `GET /service/tenantadmin/tenant-ids/{id}/availability` and expect `FREE`
- [ ] Resend an invite, then revoke the superseded original, and confirm the replacement keeps its reserved ID

### DIFF
--- a/src/main/java/de/caritas/cob/userservice/api/service/accountinvite/AccountInviteService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/accountinvite/AccountInviteService.java
@@ -368,15 +368,15 @@ public class AccountInviteService {
    * Returns the tenant ID an invite reserved to the free pool once the invite that justified the
    * reservation is durably dead (#1052).
    *
-   * <p>Tenant IDs are short, admin-visible and deliberately chosen — the Admin panel offers a
-   * "next free ID" picker — so a reservation outliving its invite silently degrades the ID space
-   * and makes that suggestion drift away from reality. {@code createInvite} already compensates
-   * this way on its failure path; revoke and expiry are the same orphan.
+   * <p>Tenant IDs are short, admin-visible and deliberately chosen — the Admin panel offers a "next
+   * free ID" picker — so a reservation outliving its invite silently degrades the ID space and
+   * makes that suggestion drift away from reality. {@code createInvite} already compensates this
+   * way on its failure path; revoke and expiry are the same orphan.
    *
    * <p>After commit, not inline: the reservation may only be freed once the terminal status is
-   * durable. Releasing inside the transaction would hand the ID back and then, on a rollback,
-   * leave a live invite pointing at an ID somebody else can now take. Outside a transaction it
-   * runs inline, which carries the same guarantee because {@code release} never throws.
+   * durable. Releasing inside the transaction would hand the ID back and then, on a rollback, leave
+   * a live invite pointing at an ID somebody else can now take. Outside a transaction it runs
+   * inline, which carries the same guarantee because {@code release} never throws.
    *
    * <p>Agency IDs are deliberately left alone. {@code agencyId} carries either a reserved ID or a
    * pre-existing agency's ID and the row does not record which, so releasing it here could free an
--- a/src/main/java/de/caritas/cob/userservice/api/service/accountinvite/AccountInviteService.java
+++ b/src/main/java/de/caritas/cob/userservice/api/service/accountinvite/AccountInviteService.java
@@ -43,6 +43,8 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.http.HttpStatus;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
+import org.springframework.transaction.support.TransactionSynchronization;
+import org.springframework.transaction.support.TransactionSynchronizationManager;
 import org.springframework.web.client.HttpClientErrorException;
 
 @Slf4j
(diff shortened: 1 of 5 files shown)


Shirloin (approved):
Approved. The eight red checks were one cause: Spotless runs before everything else, so the build died on javadoc wrapping and no test ran. Pushed the reformat — 7 lines, no code touched — and all 12 checks are green.

Revoking or expiring an invite currently burns a tenant ID forever, and since these IDs are short and admin-chosen, the "next free ID" suggestion drifts further off with every dead invite. Releasing after commit is right: freeing it inside the transaction would hand the ID back and then, on rollback, leave a live invite pointing at an ID someone else can claim.

`closingIssuesReferences` is empty despite the body saying "Closes #1052". The lost `EXPIRED` write in `acceptInvite` wants its own issue.

PR #1155, opened 2026-09-14
fix(case-handover): preserve legacy consent policy mode

shazia-k:
## Summary
- preserve `clientConsentMode` when TenantAdmin returns the legacy `clientConsentRequired` policy shape
- keep the modern `clientConsent.mode` value as the preferred source
- add a regression test for the legacy fallback

Follow-up to #1154 and its CodeRabbit review finding.

## Verification
- regression test fails before the fix (`expected: SUGGESTED but was: null`)
- `./mvnw -B test` — 4,464 tests, 0 failures/errors
- `./mvnw -B package -DskipTests` — success; `target/UserService.jar` built
- Spotless check — clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the legacy client-consent mode when the consent value is unavailable, ensuring case handover reasons remain accurate.


Shirloin (approved):
Verified locally, the new legacy-mode test is a real regession guard - restroing the old :null fallback makes it fail with "expected: <SUGGESTED> but was <null>", and it goes green again once restored.
Two extra local checks: a legacy policy set to ENFORCED also comes through as
ENFORCED (so the value is read from the stored policy, not hardcoded), and when
clientConsent.mode is ENFORCED while the legacy mode is SUGGESTED the modern
mode still wins.

Tenants whose case-handover policy is still in the legacy shape get their consent
mode back instead of null, so counsellors see whether client consent is enforced
or merely suggested before handing a case over.

Issue #1131, opened 2026-09-08
Case handover consent policy and duration are silently discarded on save


Storypapst:
## What an agency admin experiences

The Case Handover card offers a consent policy with three values — OPT_IN, OPT_OUT, NONE — and a maximum access duration. You pick OPT_OUT, you save, the card confirms. You reload, and it says OPT_IN again. The duration always reads 180 whatever you set.

Nothing warns you. The setting looks configurable and is not.

## Why

The Admin card was rebuilt with the three-value control in Admin PR #772 (merged to `pre-dev` on 2026-08-27) and PUTs its values to `/service/users/case-handover/reason-policies` (`ORISO-Admin/src/appConfig.ts:79`).

The receiving DTO on the UserService side, `CaseHandoverService.CaseHandoverReason` (`ORISO-UserService/src/main/java/.../service/CaseHandoverService.java:1138-1147`), has **neither** `clientConsent` **nor** `maxAccessDurationMinutes`. Spring drops unknown properties silently, and the GET never returns them either. So `resolvedClientConsentPolicy` (`ORISO-Admin/src/.../caseHandoverCardUtils.ts:70-74`) always falls back to `clientConsentRequired ? 'OPT_IN' : 'NONE'`.

Measured on `origin/dev` (`10a29ac0`) and `origin/pre-dev` on 2026-09-08. Both affected.

## Why this matters beyond a cosmetic bug

The setting governs whether an advice seeker is asked before another counsellor of the same agency reads along. An agency that believes it has switched to "allowed by default, may object" is in fact still on "ask every time" — or the other way round, depending on the boolean underneath. A configuration screen that silently discards a data-protection setting is worse than one that does not offer it.

## Two options

1. **Extend the DTO** and persist both fields, which makes the control honest. This is the direction the requirement document assumes.
2. **Disable the control** until the backend carries the values, so nobody configures something that is not stored.

Option 2 is the smaller change and restores honesty immediately; option 1 is the actual goal.

## How should I test this?

- [ ] As a tenant admin, set a handover reason's consent policy to OPT_OUT and save. Reload the page: the value is still OPT_OUT.
- [ ] Set a maximum access duration other than 180, save, reload: the value persists.
- [ ] Set the policy to NONE, save, reload: still NONE.
- [ ] Confirm the value actually reaches the database, not only the form.
- [ ] Confirm an advice seeker's experience changes accordingly for the chosen mode.
[...]

PR #1165, opened 2026-09-15
fix(case-handover): let the typed consent policy win over the echoed mode

Storypapst:
## What still breaks

#1131 is half fixed. #1148/#1154/#1155 made the backend carry `clientConsent` and `maxAccessDurationMinutes` — that part works. Two gaps remained; this PR closes the backend one.

The Admin re-sends the whole reason list it last read, so every entry still carries the `clientConsentMode` string from the previous GET — positioned **after** `clientConsent`, because that is the DTO's field order. Jackson applied the policy object first and the stale string second:

```json
{"clientConsent": {"value": "OPT_OUT", "mode": "ENFORCED"},
 "clientConsentMode": "SUGGESTED"}       ← wins today
```

So picking **"Zustimmung Opt-Out (Vorgabe)"** stores **"Zustimmung Opt-Out (anpassbar)"**. The value lands, the mode silently degrades — for a data-protection setting the mode is the half that says whether an agency may override it.

## What changes

- The typed policy object's `mode` wins over a separately sent `clientConsentMode`, in either field order. The loose string stays the fallback for callers that do not send an object (Frontend, legacy).
- No API path, no contract, no migration. The response still emits the bare string plus `clientConsentMode`, so the Frontend (`apiCaseHandover.ts`) is untouched.

Also pins the #1131 acceptance case that had no test at all: a changed `maxAccessDurationMinutes` reaches the TenantService write-through *and* comes back on the same response the card re-reads.

## The other half

The remaining symptom is in the Admin, not here: the card writes the policy object but reads it back expecting an object, while this service answers the bare string. `resolvedClientConsentPolicy` never normalised, so every stored value looked discarded after a reload. Fixed in OpenResilienceInitiative/ORISO-Admin#970 — the two belong together for #1131 to be testable end to end.

## Verification

JDK 21, local:

- `CaseHandoverReasonJsonTest` — 2 new wire-order cases; the first fails on `dev` with `expected: <ENFORCED> but was: <SUGGESTED>`.
- `CaseHandoverServiceTest` — 104/104, including the new duration round trip (written policy carries 90, response carries 90).
- Full suite `./mvnw -B test` — **4,495 tests, 0 failures, 0 errors**, BUILD SUCCESS.
- `spotless:apply` clean.

## Reviewer test plan

- [ ] `./mvnw test -Dtest='CaseHandoverReasonJsonTest,CaseHandoverServiceTest'` → green.
- [ ] Drop the `consentModeFromPolicyObject` guard in `setClientConsentMode` → `typedPolicyModeWinsOverTheStaleModeEchoedBackByTheAdmin` fails. Confirms the test is load-bearing.
- [ ] On Dev, with the Admin PR deployed: set "Zustimmung Ratsuchende Person" to **Opt-Out (Vorgabe)** on "Rat benötigt", save, reload → still Opt-Out **and** still "Vorgabe" (not "Empfehlung").
- [ ] Set the max session duration to 1 h 30 min, save, reload → still 1 h 30 min.
- [ ] Frontend: `/service/users/case-handover/reasons` still returns `clientConsent` as a string.

Not covered here, and it needs a live tenant admin: the issue's "confirm the value actually reaches the database". The write-through is proven against a mocked TenantService client only.

Refs #1131. Related: #1148, #1154, #1155, OpenResilienceInitiative/ORISO-Admin#771.

## Summary by CodeRabbit

- **New Features**
  - Case-handover policy updates now support saving a requested access duration when no duration policy exists.
  - Typed consent settings now take precedence over conflicting legacy consent-mode values.
  - Updated policy responses include the configured consent and duration settings.

- **Bug Fixes**
  - Corrected handling of consent modes for typed policy input and standalone consent values.


Storypapst:
### A second bug, found by moving the test to the wire

The tests in this PR so far asserted on a captured argument — `verify(caseHandoverPolicyCacheService).updateEffective(eq(7L), captor)`. A Mockito capture cannot disagree with the code that built it. So I added a test at the real seam: `MockRestServiceServer` on the generated client's `RestTemplate`, the Admin's raw JSON in through the runtime mapper, the assertion on the **HTTP body that actually leaves this service**. Only the socket, the identity provider and the cache store are stubbed.

It went red immediately, on a bug neither this PR nor #1148/#1154/#1155 had caught:

```
expected: 90
 but was: (absent)   ← maxAccessDurationMinutes never reached the wire
```

`applyPolicyUpdate` wrote the duration only when the stored reason **already** carried a `maxAccessDurationMinutes` policy object:

```java
&& policy.getMaxAccessDurationMinutes() != null
```

TenantService's `resolveReason` returns null for that field whenever the parent carries none — the state of any tenant seeded before the duration policy existed. For those tenants the admin's duration was dropped before the request was even built. No error, no log, no 400. **That is the "duration always reads 180" half of #1131** — the half I had reported as covered, on the strength of a mocked test.

Fix: create the tenant-local policy at the platform default mode for that field when it is missing. Commit `6e03a5c6`.

### Verification

- New `CaseHandoverPolicyHttpContractTest` — red before the fix with the message above, green after.
- Full suite `./mvnw -B test`: **4,496 tests, 0 failures**, BUILD SUCCESS. Spotless clean.

### One more reviewer step

- [ ] Revert the `durationPolicy == null` branch in `applyPolicyUpdate` → `aChangedDurationReachesTenantServiceEvenWhenTheStoredReasonCarriesNoDurationYet` fails. Confirms the test pins the fix.

Shirloin (approved):
Verified locally: the chosen consent mode now survives the Admin's save, and a changed duration reaches TenantService even for tenants without a stored duration policy. The case handover tests pass (112/112), and both new tests fail when their fix is removed.

Issue #1168, opened 2026-09-16
Admin/consultant profile updates wipe Keycloak user attributes (userId) → agency admins locked out with 403


Storypapst:
## What happens

A Beratungsstellen-Admin (`restricted-agency-admin`) is bounced to `/admin/access-denied` on every Admin login as soon as their admin record has been edited once in the Admin UI. The page has no logout, so the account is stuck until the session expires.

Reproduced on Dev (dev.oriso.org) on 2026-09-16 with a freshly created agency admin. Ingress log of that session:

```
06:12:14  POST /service/useradmin/agencyadmins                                   200   ← admin created
06:12:23  PUT  /service/useradmin/agencyadmins/8ed43c2c-…                        200   ← admin edited
06:29:34  POST /service/useradmin/admins/8ed43c2c-…/grant-consultant-identity   200
06:31:54  GET  /service/agencyadmin/agencies?q=*&page=1&…                       403   ← every login from here on
```

AgencyService log for each 403:

```
AccessDeniedException: Domain user id is required for this operation
  at de.caritas.cob.agencyservice.api.util.AuthenticatedUser.requireUserId(AuthenticatedUser.java:64)
```

## Root cause

`KeycloakService.updateProfile` (and `updateDummyEmail`) build a **fresh** `UserRepresentation` from the profile values (`getUserRepresentation(profile)`) and `PUT` it to Keycloak. That representation carries an attribute map with only `username`/`userName`/`tenantId`. Keycloak replaces the whole attribute map when one is sent, so every attribute creation wrote and the update did not know about is dropped — above all **`userId`**, the attribute behind the custom `userId` token claim, and `locale`.

`UpdateAdminService.updateAgencyAdmin` (PUT `/useradmin/agencyadmins/{id}`) and `ConsultantUpdateService` (PUT `/useradmin/consultants/{id}`) both go through `updateProfile`, so **every admin or consultant edited in the Admin UI loses the claim**. The AgencyService needs that claim to scope a restricted agency admin to their own agencies (`AgencyAdminSearchService.agencyAdminFilterPredicate` → `requireUserId()`, hardened in OpenResilienceInitiative/ORISO-AgencyService#143), hence the 403 on the very first landing request, which the Admin's `fetchData` turns into the global `/admin/access-denied` redirect.

## Fix (three small PRs)

[...]

PR #1169, opened 2026-09-16
fix(keycloak): keep existing user attributes on profile updates (#1168)

Storypapst:
Fixes OpenResilienceInitiative/ORISO-UserService#1168 (root cause).

## Problem

`KeycloakService.updateProfile` and `updateDummyEmail` PUT a freshly built `UserRepresentation` whose attribute map only holds `username`/`userName`/`tenantId`. Keycloak replaces the whole attribute map when one is sent, so every admin or consultant edit in the Admin UI drops the attributes creation wrote — above all `userId` (the custom token claim the AgencyService scopes restricted agency admins with) and `locale`. After the first edit a Beratungsstellen-Admin gets `403 Domain user id is required` on their agency list and the Admin UI bounces them to `/admin/access-denied` on every login. Reproduced on Dev 2026-09-16, timeline and logs in the issue.

## Change

- `updateProfile` / `updateDummyEmail` merge the profile attributes onto the attributes Keycloak currently holds instead of replacing them (`withExistingAttributes`).
- No behaviour change for the fields the update did set; existing tests unchanged.

## Tests

- New: `updateProfile_Should_preserveAttributesKeycloakAlreadyHolds`, `updateDummyEmail_Should_preserveAttributesKeycloakAlreadyHolds`.
- `KeycloakServiceTest`: 107 run, 0 failures (local, JDK 21; the suite needs JDK 21 locally because Mockito inline mocks fail on 25).
- Verified locally only — not deployed anywhere.

## Reviewer test plan

1. On Dev, create an agency admin, then edit them once (name/email) in the Admin UI.
2. Log in as that admin → the agency list must load (before this fix: 403 → access-denied loop).
3. In Keycloak, the user's `userId` and `locale` attributes must still be present after the edit.

Companion PRs: AgencyService fallback to the token subject, Admin logout on the access-denied page (links in the issue).

## Summary by CodeRabbit

- **Bug Fixes**
  - User profile, email, language, deactivation, and account updates now preserve existing account attributes.
  - Language preferences are updated correctly even when no previous attributes are available.
  - Administrative and consultant account changes retain existing identity information, including across tenant-related updates.
  - Updates now handle missing or unreadable account data more reliably without losing stored information.


Storypapst:
Verified on Pre-Dev together with ORISO-AgencyService#292 — baseline reproduced with the dev images (userId claim wiped by the edit, agency list 403), fixed images: claim survives the edit, list 200. Details: https://github.com/OpenResilienceInitiative/ORISO-UserService/issues/1168#issuecomment-5696525468

shazia-k (approved):

Issue #1181, opened 2026-09-17
Read invitation context and suggest guest names before creating an account


Storypapst:
**Why**
Opening a Live Chat invitation currently creates an account before the person chooses a name. Live Chat and mail registration also obtain names through different paths. The approved shared-entry plan needs a common source of matching names and animal avatars, and a way to read an invitation before joining.

Parent: OpenResilienceInitiative/ORISO-Frontend#1216. This is the additive backend prerequisite; changing the join flow and switching clients are separate packages.

**What to build**
Provide bounded name suggestions and public invitation context without creating users, reserving names, changing invitation status or adding waiting entries. Check names consistently across the existing identity systems. A dependency outage must return a retryable failure, never claim that an unchecked name is free. Preserve existing clients and the current redeem contract.

**Acceptance**
- [ ] Repeatedly opening context and rolling suggestions leaves accounts, invitations and waiting entries unchanged.
- [ ] Suggestions contain a coherent username, display name and existing animal-avatar key; requested exclusions and bounded work are enforced.
- [ ] Name conflicts and unavailable identity systems produce distinguishable outcomes.
- [ ] Inactive, expired and missing invitations return an appropriate public error without mutating the invitation.
- [ ] Unauthenticated HTTP access, tenant boundaries and existing public-request protections are tested through real application wiring.
- [ ] Existing redemption and legacy anonymous accounts continue to work.

### Reviewer test plan
- [ ] Read a valid invitation and request four names repeatedly → no account or waiting place appears, and each name has the matching animal.
- [ ] Replace one proposal while excluding the other three → exactly one distinct proposal is returned.
- [ ] Make an identity provider unavailable → the request fails as retryable rather than returning an unchecked available name.
- [ ] Read an expired invitation → a local error appears and the link record is unchanged.
- [ ] Use the existing redeem endpoint → its response and account/session behavior remain compatible.

**Affected repositories and delivery**
- Backend: OpenResilienceInitiative/ORISO-UserService#1182.
- Required public endpoint cap: OpenResilienceInitiative/ORISO-Helm#355; deploy before public API activation.
- Both PRs target `dev`. Frontend cutover and explicit idempotent Join remain separate packages.

Storypapst:
**Short thesis:** The additive backend reads are implemented and locally verified; public exposure also requires the narrow Helm rate cap.

[...]

Issue #355, opened 2026-07-09
test: add unit tests for CreateConsultantDTO, CreateAnonymousEnquiryResponseDTO, UpdateConsultantDTO, TwoFactorAuthDTO, AppintmentEnquiryData, EnquiryData, ConsultantResponseDTO


Kadyan25:
## What
Add unit test coverage for the following classes:

- `CreateConsultantDTO` — generated OpenAPI DTO, consultant creation payload (452 missed instructions, 5% current)
- `CreateAnonymousEnquiryResponseDTO` — generated OpenAPI DTO, anonymous enquiry response (341 missed, 12% current)
- `UpdateConsultantDTO` — generated OpenAPI DTO, consultant update payload (272 missed, 17% current)
- `TwoFactorAuthDTO` — generated OpenAPI DTO, 2FA setup fields (290 missed, 0% current)
- `AppintmentEnquiryData` — Lombok model, appointment enquiry payload
- `EnquiryData` — Lombok model, enquiry payload with consultant email
- `ConsultantResponseDTO` — generated OpenAPI DTO, consultant response payload

## Scope
- Tests only — no production code changes (bugs found → fix in same PR per Frank's rule)
- New tests added below existing ones in each file with a separator comment
- All classes covered in one PR

## Verification
- [x] All new @Test methods pass — 70 tests across 7 classes
- [x] No existing tests were modified or removed
- [x] 80%+ instruction and branch coverage confirmed via JaCoCo — all 7 classes at 99.3-100% instruction / 95-100% branch

Kadyan25:
PR opened: https://github.com/OpenResilienceInitiative/ORISO-UserService/pull/368

Storypapst:
Resolved via PR#368 (merged 2026-07-11). Closing as part of the GitHub hygiene sweep — issue was already fixed but never marked closed.

PR #1182, opened 2026-09-17
feat: read guest names and invitation context before joining

Storypapst:
## What happened

The invitation page currently creates a guest account before the person chooses a name. This first refactoring slice lets the page read invitation context and obtain four checked name/avatar suggestions without creating an account or session. Existing redeem and registration behavior stays compatible until the frontend cutover.

Refs OpenResilienceInitiative/ORISO-UserService#1181
Parent: OpenResilienceInitiative/ORISO-Frontend#1216

## What changed

- Add public, read-only `GET /users/invitelinks/{token}/context` and `POST /users/identity-suggestions`. A request asks for one or four candidates and may exclude up to twenty prior usernames.
- Move the existing curated animal/name catalogue into a backend-owned resource. Each result keeps its actual username, display name and existing avatar asset together. No reservation is made; joining must recheck availability in the later slice.
- Check the global user/consultant, identity-provider and Matrix namespaces. Unknown availability returns 503 instead of treating a failed lookup as an available name. Legacy Matrix callers retain their behavior; the new strict lookup uses an externally provisioned admin token and never invokes login or account bootstrap.
- Preserve the complete caller tenant context, including an absent tenant ID. The new endpoints never provision credentials or write invitation status.
- Limit input and candidate attempts; expose only the two exact HTTP routes through the existing public security configuration.

## Verification

Local only on Java 21; no deployment or browser acceptance of these new endpoints yet. Earlier baseline full unit suite: 4,675 tests passed without failures or skips; that full suite was not rerun for the later review corrections. Package/repackage and the Maven formatting gate passed. The package lifecycle reran the unit suite because this POM uses `skip.unit-tests`, not the generic `skipTests` property. Targeted suite: 98 tests passed, including real controller/service/security wiring with external dependencies isolated. Tests exercise invalid/expired invites, collisions, dependency failure, caller-context restoration and legacy redeem regression behavior.

The first new context compile probe failed because the new API did not exist; strict Matrix tests were authored before implementation, but their initial red execution was not independently captured. No broader TDD claim is made.

CI follow-up: the architecture check caught a direct adapter import in the new identity service. Commit `40c3c20b` routes strict Matrix checks through the existing `MatrixUserClient` port. All 100 Python CI contracts, 89 targeted Java tests for affected callers and HTTP wiring, formatting and package/repackage pass locally. Those checks passed at that earlier head.

## Review corrections

The cold-cache HTTP regression exposed an unintended Matrix admin login behind the previous token helper. Strict availability now uses only `MATRIX_AVAILABILITY_ADMIN_ACCESS_TOKEN`; missing or rejected credentials return sanitized 503 without fallback. Companion Helm PR #355 supplies an explicit existing-Secret reference and rotation/activation instructions. This is an admin credential used only for GET by this code, not a scoped read-only token.

Dependency diagnostics retain only the exception class, never the original message or cause. Deterministic catalogue tests cover Cyrillic fallback, exact truncation and separator cleanup; the exclusion-size fixture now uses valid names so it tests the intended bound.

Current local verification: 135 targeted Java tests, 109 Python CI/OpenAPI tests plus two subtests, Maven formatting and package/repackage all pass. HTTP tests first reproduced the unwanted login; the diagnostic test first reproduced the missing safe log. Independent read-only review found no further concrete auth/Secret-reference issue. No deployment, runtime credential provisioning or browser acceptance has occurred.

## How should I test this?

- [ ] Request context for an active invitation → receive tenant/topic/entry metadata, no credentials and no new account/session.
- [ ] Request context for an invalid, inactive or expired invitation → receive the documented 4xx and no writes.
- [ ] Request four German suggestions → receive four distinct valid usernames with matching existing avatar keys.
- [ ] Request one replacement excluding the other cards → receive a different candidate; prior cards need no reload.
- [ ] Make an identity dependency unavailable → receive 503, never a supposedly available name.
- [ ] Start a fresh UserService with the configured lookup token → Matrix receives only existence GETs from suggestions, no login or registration.
- [ ] Remove or revoke that token → suggestions return 503 without falling back to admin username/password. Restore/rotate the Secret and restart → valid lookups recover.
- [ ] Exercise the existing invitation redeem flow → retain its current behavior.

## Rollout boundary

Deploy the companion endpoint rate limit (OpenResilienceInitiative/ORISO-Helm#355) and provision its referenced Matrix lookup token before activating the new entry UI. Missing credentials intentionally make suggestions unavailable; legacy clients retain their current paths. This PR alone does not change the frontend, defer account creation to Join, implement idempotent Join, fix the reported Staging freeze, or alter lifecycle timers. Those remain separate plan packages.

## Summary by CodeRabbit

* **New Features**
  * Added guest identity suggestions with localized names, animal avatars, availability checks, exclusions, and duplicate prevention.
  * Added a read-only endpoint for viewing invitation-link context without redeeming the link.
  * Added strict username availability checks with clear service-unavailable responses when verification is inconclusive.
  * Added localized guest-name catalogues for German, English, French, Spanish, Russian, and Ukrainian.
* **Security & Reliability**
  * Enabled anonymous access for guest suggestions and invitation context while preserving CSRF and caching protections.
  * Improved sanitization of upstream error details and tenant-context handling.


Shirloin (dismissed):
Tested: merged current dev (8d7295c7, 49 commits ahead of this branch) without conflicts. The guest/identity, Matrix strict-availability, Keycloak and invite suites pass on the merged state (244 tests). Turning a dependency failure into "available" fails 4 GuestUsernameAvailability tests, so fail-closed is pinned. Only the two new routes are permitAll, and the context response carries IDs only. Not deployed.

Merge together with Helm #355, which provides the rate-limited ingress CodeRabbit asked for before these routes go public.

shazia-k (dismissed):

shazia-k (approved):

PR #1184, opened 2026-09-17
Serve canonical contract mail previews with tenant and invitation guards

Storypapst:
The contract-forwarding dialog needs to show the email that will actually be sent. The existing generic invitation preview uses a different renderer. These read-only endpoints ask ConsultingTypeService's canonical signing-mail renderer for the preview, using an explicitly non-deliverable sample link and expiry.

Refs OpenResilienceInitiative/ORISO-Admin#723
Refs OpenResilienceInitiative/ORISO-Admin#951

**What changed**

- Authenticated `POST /useradmin/dpa-invites/preview` accepts a tenant ID. Tenant administrators can preview only their own tenant; explicit platform administrators can select another tenant. Null context fails closed.
- Public `GET /users/account-invites/{token}/onboarding/dpa-mail-preview` accepts only an active, unexpired tenant-admin invitation. The tenant comes from the invitation, never a caller-provided ID.
- Both return `{subject, html}` from CTS. They do not mint a signing link, persist onboarding progress, consume the invitation or send an email.
- Only the exact public GET route is added to anonymous access. The generic template-manager preview is unchanged.

**Dependency and rollout**

Requires OpenResilienceInitiative/ORISO-ConsultingTypeService#147 to be deployed first. The companion Admin change for #723 follows these additive endpoints. Keep draft until companion PR links, CI and review are complete.

**Validation**

Local only, JDK 21. No deployed image, real email receipt or authenticated Dev acceptance yet.

- Full Maven package passed; Surefire XML records 500 suites, 4,664 tests, zero failures/errors/skips.
- Focused coverage includes same/foreign tenant, platform cross-tenant access, null claims, active/expired/revoked/accepted/wrong-role/unknown invitations, and canonical downstream rendering without sending.
- Spotless passed; `git diff --check` passed.
- Supplemental `mvn checkstyle:check` failed with 22,387 repository-wide diagnostics, including Javadoc/line-format diagnostics in new files. This raw goal is not configured as the repository's formatting gate; the configured Spotless gate passed. No claim of a clean Checkstyle run is made.

### Reviewer test plan

- [ ] As tenant admin, open contract forwarding → see your tenant's email preview; request a different tenant's preview → receive 403 before rendering.
- [ ] As platform admin, select a tenant and open forwarding → see that tenant's preview.
- [ ] Open a fresh tenant-admin invitation and choose forwarding → preview loads before registration or login.
[...]

PR #1188, opened 2026-09-18
fix: make the DPA mail preview work in the onboarding wizard

Storypapst:
## Problem

The forward-to-authorised-signer dialog shows **"Die Vorschau konnte nicht erzeugt werden"** instead of the mail, on Dev, for every real onboarding.

Since ORISO-Admin `0d083502` (17 Sep) the dialog no longer renders the mail locally — it fetches the canonical document from UserService (#1184). Any non-2xx now surfaces as that error box; before the change the preview could not fail at all. That is why this looks like a regression.

## Root cause: the status guard contradicts the wizard's own lifecycle

`validatePreviewableTenantAdminInvite` required the invite to be `EMAIL_SENT`.

But the DPA step runs **after** account registration, and `registerTenantAdmin` claims the invite through `claimForAcceptance`, whose `UPDATE` sets `status = ACCEPTED`. The two-factor guard in `TenantAdminOnboardingService` states the same lifecycle from the other side:

```java
if (invite.getStatus() == AccountInviteStatus.EMAIL_SENT) {
  throw new BadRequestException("Registration has not happened yet for this invite");
}
if (invite.getStatus() != AccountInviteStatus.ACCEPTED) { throw linkDeathException(invite); }
```

So by the time the forward dialog can be opened, `ACCEPTED` is the ordinary state and `EMAIL_SENT` would mean "registration has not happened". The preview demanded exactly the state that cannot occur there, answered **400 "Account invite is not active"**, and the dialog fell into its error branch — deterministically, for every onboarding.

The existing tests did not catch this because one of them actively pinned the wrong expectation (`preview_consumedInvite_rejectsWithoutCallingCanonicalRenderer`).

### Change

`PREVIEWABLE_STATUSES = {EMAIL_SENT, ACCEPTED}`. Terminal links (revoked, superseded, expired) stay refused, and the expiry check is unchanged. Rendering remains read-only — it mints no sign link and sends no mail — so widening this state does not widen what a token can cause.

## Second, independent finding: the gateway-prefixed path is not mapped

While tracing the 404 space I found that both DPA endpoints map only the un-prefixed path, while every neighbouring onboarding endpoint maps **both** (`/users/…` and `/service/users/…`):

- `TenantAdminDpaMailPreviewController` — `@GetMapping` now lists both.
- `DpaForwardEmailController` — class-level `@RequestMapping` now lists both, covering `/preview` and `/email`.

On Dev the `userservice-rewrite-ingress` strips `/service`, so this is not what caused the reported bug. It does bite when the frontend runs against a local backend: the Vite dev proxy forwards `^/(service|api/v1|auth)` **unchanged**. `UserController` and `UserAdminController` already map both deliberately "without relying on the gateway to strip the `/service` prefix" — this follows that decision. `SecurityConfig` line 493 already grants `/service/useradmin/**` the same authorities, and both handlers keep `@PreAuthorize(ADMIN_AUTH)`, so nothing becomes more permissive.

## Tests

The pre-existing preview tests call the handler method directly, so they could see neither the routing gap nor the lifecycle mismatch. Three new tests:

- `preview_registeredInviteAtTheDpaStep_rendersTheMail` — fails with `BadRequest: Account invite is not active` before the guard fix.
- `preview_isServedOnTheGatewayPathTheAdminFrontendCalls` — MockMvc, fails `expected:<200> but was:<404>` before the mapping fix.
- `previewSigningMail_isServedOnTheGatewayPathTheAdminFrontendCalls` — same for the admin endpoint.

## Verification

**Tested locally only** — not yet deployed or verified on an environment. `mvn test` on Java 21 for `TenantAdminDpaMailPreviewControllerTest`, `DpaForwardEmailControllerTest`, `TenantAdminOnboardingControllerTest`, `AccountInviteControllerTest`: 70 tests green. Spotless applied.

Evidence from Dev supporting the diagnosis: an unmapped path under the same prefix answers `401`, while `/service/users/account-invites/{token}/onboarding/dpa-mail-preview` answers `404` for an unknown token — the same signature as the working neighbouring endpoint. So the endpoint is deployed and reachable on Dev, and the failure a real token hits is the status guard, not routing.

## Reviewer test plan

1. On Dev, run a tenant-admin onboarding through the account and 2FA steps.
2. At the DPA step choose "Ich bin nicht unterschriftsberechtigt".
3. The mail preview renders; the call to `…/onboarding/dpa-mail-preview` returns 200 instead of 400.
4. Confirm a revoked or expired invite still gets 400.

## Note for the deploy side (not part of this PR)

`UserService Build and Publish` has failed on every `dev` merge since at least 16 Sep, in the `Reject fixable high or critical image vulnerabilities` step (Trivy). The image is pushed *before* that gate, so `:dev` is current — but the pipeline has been red for days and nobody gets a green signal. Worth a separate look, cc @Hassan9215.

Refs #723, #951


Shirloin (commented):
The premise doesn't match the wizard order. In Admin `useTenantAdminOnboardingFlow.ts` the forward dialog only opens in the `organisation` phase, before `registerTenantAdmin` runs, so the invite is still EMAIL_SENT when the preview is requested. `registerTenantAdmin` itself requires the forward to exist already, and `forwardDpa` (`reserveDpaForwardAttempt`) also only accepts EMAIL_SENT. Widening the preview guard to ACCEPTED therefore doesn't change what a real onboarding hits.

Could you capture the actual status and body of `…/onboarding/dpa-mail-preview` on Dev while the dialog shows the error? One candidate in the same guard: `expiresAt == null` also answers 400 "Account invite is not active". Another is a failure inside `previewSigningMail`.

The `/service/...` mappings are fine to keep: both paths are already in SecurityConfig with the same authorities.

